PR #18, opened 2017-08-29
Fix bug in argument passing

quixoten:
Because the Fiber block does not use splat:

```
@fiber = Fiber.new do |args|
...
end
```

only one argument was being passed from the `#run` method in the call to
`@fiber.resume(*args)`. This fixes the issue by removing the splat from
the call to `resume`. This could also be fixed by adding a splat to the
block parameter. I chose this method under the assumption that it will
be faster by skipping the parameterization step when it's not necessary,
but adding the splat to the block parameter might be more consistent
with the rest of the code.

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -113,7 +113,7 @@ module Async
 		def run(*args)
 			if @status == :initialized
 				@status = :running
-				@fiber.resume(*args)
+				@fiber.resume(args)
 			else
 				raise RuntimeError, "Task already running!"
 			end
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks for this, and thanks for the very clean PR with initial failing test case.

ioquatix:
For some reason this is broken on JRuby.

https://travis-ci.org/socketry/async/jobs/273114135

If you have time do you think you can take a look? Otherwise if not, let me know and I'll try to make some time.

quixoten:
interesting! yeah, i'll take a look.

PR #19, opened 2017-08-31
Improve code readability in README

jgaskins:
The tabs display as 8-space indentation in the browser. This PR converts them to 2-space indents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,9 @@ The highest level entry point is `Async::Reactor.run`. It's useful if you are bu
 
 ```ruby
 def run_server
-	Async::Reactor.run do |task|
-		# ... acccept connections
-	end
+  Async::Reactor.run do |task|
+    # ... accept connections
+  end
 end
 ```
 
@@ -72,9 +72,9 @@ If you can guarantee you are running within a task, and have access to it (e.g. 
 
 ```ruby
 def do_request(task: Task.current)
-	task.async do
-		# ... do some actual work
-	end
+  task.async do
+    # ... do some actual work
+  end
 end
 ```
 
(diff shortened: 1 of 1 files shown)


tarcieri:
The question of tabs strikes again.

@ioquatix this is just going to keep coming up until you give in 😜

jgaskins:
lol! I admit that I didn't even check the commit history or closed PRs before submitting this. To be honest, I personally subscribe to the [Aaron Patterson school of thought regarding tabs v spaces](https://twitter.com/tenderlove/status/849454313575514112), but project `README`s are almost exclusively consumed via GitHub, whose Markdown preview doesn't support sanitization of tabs into spaces AFAIK.

Since the Ruby convention is 2 spaces and Ruby has [the most consistent programming-language ecosystem regarding tabs/spaces](https://ukupat.github.io/tabs-or-spaces/), this feels to me like a sensible change. :-)

picatz:
This is hilarious.

picatz:
_Also_: I totally support this PR being merged. I was thinking about doing this myself the other day while looking at it 😹

ioquatix:
I'd rather GitHub fix their rendering to honour `.editorconfig` than work around the issue. https://github.com/socketry/async/blob/master/.editorconfig

picatz:
Where do you even make that PR though?

ioquatix:
GitHub doesn't have an official place to report issues, except via the support email. Although, I've had a pretty good response working directly with GitHub developers to fix issues.

picatz:
> @ioquatix : GitHub doesn't have an official place to report issues


That's like... _almost_ ironic. 😹

ioquatix:
Yeah, there has been some attempts to make something, here is one of the more popular ones: https://github.com/isaacs/github but totally unofficial.

tarcieri:
@ioquatix until GitHub fixes the problem (if that ever even happens) perhaps it's worth taking the simpler measure of adjusting the whitespace so it isn't immediately off-putting to the overwhelming majority of Rubyists who look at it and find it off-putting?

Is the number of times this has come up not evidence that it's a problem? I think it's pretty clear I'm not the only one who looks at it and finds it to be "weird"

ioquatix:
I review - indentation may look odd to some, but it looks fine to me, it doesn't cause any line wrapping which would be more of a concern, I believe the solution here is for github to fix and respect the `.editorconfig`.

PR #20, opened 2017-09-08
Pass args in a way that works in JRuby too

quixoten:
Fixes: https://github.com/socketry/async/pull/18#issuecomment-327977891


ioquatix:
Thanks for taking the time to look into this and find a solution. May Ruby be with you.

PR #21, opened 2017-09-20
fixes: code examples in README use tabs for identing

krisleech:


krisleech:
Sorry I didn't see this had already been discussed https://github.com/socketry/async/pull/19.

I'd :+1: having spaces instead of tabs, for me first contact with README is always via Github. After that text editor which looks like this

ioquatix:
I've flicked off another email to GitHub support to see what the status of this is.

> Hello.
> 
> Even though I have an .editorconfig which specifies tab with logical size 2 (or 4) spaces, it’s not respected in code in the READMD.md.
> 
> Here is editor config file: https://github.com/socketry/async/blob/master/.editorconfig
> 
> Here is readme: https://github.com/socketry/async
> 
> Here is normal code: https://github.com/socketry/async/blob/master/lib/async/task.rb
> 
> It seems to be a bug that the .editorconfig is not respected by the README.md markdown renderer?
> 
> Thanks
> Samuel

As you can see, .editorconfig is correctly used in code, just not in `README.md`. This is really a bug with GitHub and I do agree 8-space tab width can be a bit jarring, although none of the code in the README is wrapping because of that, it's just a bit bigger than usual.

Issue #23, opened 2018-07-04
Use Fiber.transfer instead of Fiber.yield


funny-falcon:
`Fiber.yield` is used by Enumerator, so if enumerator will call to some io, it will return to calling fiber instead of being scheduled.
But if scheduler will use `fiber.transfer`, than it will play nicely with Enumerator.

ioquatix:
I will probably work on this in the next week, but if you are interested, do you think you can make a failing test case to demonstrate the problem? I got the general idea from Ruby bug tracker.

funny-falcon:
I will try.

ioquatix:
Don't try to fix this issue as it's pretty deeply embedded in the current API, but if you can provide a failing test case using enumerator it gives us a goal to work towards.

funny-falcon:
I was not quite right: Enumerator uses Fiber.yield only in "external iterator" mode.
Still `zip` method uses this mode, for example.
External iterator mode could be used for merging two sequences (ie merge-sort, or other kind of merges).
And some times, it is desirable to use Fiber.resume+Fiber.yield in user code directly.

I added test in #24

ioquatix:
Thank you so much this is awesome!

ioquatix:
I will make separate branch for this. There might be some limitations of transfer because it has a direct relationship between reactor and task, but in some cases the relationship is a bit more complex.

https://github.com/socketry/async/tree/fiber-transfer

ioquatix:
Okay, so I've played around with this a bit.

There are some areas where it might be feasible to use `#transfer`, and I'll try to get your specs passing.

However, now that I think about it, why doesn't `Enumerator` use `transfer`? Because it has all the state, it should be trivial to do so, right? Then, this is a non-issue because normal `Fiber.yield` can't affect it? I'd have to play with the implementation to say for certain, but I wonder if you thought about it.

ioquatix:
Here is an example of why this might be impossible:

```
Async.run do |parent| # Fiber.new.resume
	parent.async do # Fiber.new.resume
		# Two kind of implementation:
		reactor.sleep # Implemented by Fiber.yield -> back to parent
		reactor.sleep # Implemented by Reactor.transfer -> back to reactor, parent never resumed.
		
		# Another example:
		input.read # Implemented by Fiber.yield -> back to parent
		input.read # Implemented by Reactor.transfer -> back to reactor, parent never resumed.
	end
end
```

[...]

PR #24, opened 2018-07-05
Test for Enumerator as external iterator, and explicit Fiber communication

funny-falcon:
While external iteration is not very useful, it could unintentionally appear with `zip` method.
Also, no one prevents from using Fiber.yield+Fiber.resume directly. It is sometimes really useful.

Test for #23


ioquatix:
I will merge this and experiment.

PR #34, opened 2019-03-03
Warn user that closing selector may cause IO leak

jeremyjung:
Gives the user a hint that closing a selector with active monitors may cause their socket or file descriptor to leak.  

Do you think a specific error like `LeakError` should be added?  If so, where would you define it?

### DIFF
--- a/lib/async/debug/selector.rb
+++ b/lib/async/debug/selector.rb
@@ -67,7 +67,7 @@ module Async
 			
 			def close
 				if @monitors.any?
-					raise RuntimeError, "Trying to close selector with active monitors: #{@monitors.values.inspect}!"
+					raise RuntimeError, "Trying to close selector with active monitors: #{@monitors.values.inspect}! This may cause your socket or file descriptor to leak."
 				end
 			ensure
 				@selector.close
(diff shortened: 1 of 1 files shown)


ioquatix:
It's not an error you'd expect the user to catch since it's only during debug operation.

That being said, you could add

```ruby
class DescriptorLeakError < RuntimeError
	def initialize(wrappers)
		super "Trying to close selector with active monitors: #{wrappers.inspect}! This may cause your socket or file descriptor to leak."
	end
end
```

to the top of that file and use that.

ioquatix:
Maybe just call it `LeakError`.

lib/async/debug/selector.rb
```diff
@@ -67,7 +73,7 @@ def wakeup
 			
 			def close
 				if @monitors.any?
-					raise RuntimeError, "Trying to close selector with active monitors: #{@monitors.values.inspect}! This may cause your socket or file descriptor to leak."
+					raise LeakError
```
ioquatix: You need to pass in `@monitors.values`, i.e. `raise LeakError, @monitors.values`.

jeremyjung: Oops, my bad. 😅

ioquatix:
Thanks for your work.

PR #35, opened 2019-03-06
Unblock fibers in order when waiting for enqueue on LimitedQueue

Bringer128:
First, just wanted to say this is an amazing library!

I was playing with LimitedQueue and I was surprised when I tried to enqueue stuff from multiple tasks they would all get resumed on my first dequeue. I was expecting the enqueuing tasks to be resumed in the order they invoked enqueue.

Now possibly this is mismatched expectations as I was trying to use this Queue with multiple producers. If this isn't the intention and this library only wants to support single producer-consumer use cases then we can close this PR.

The fix summary: I added an inner queue to the `LimitedQueue` that works in reverse to the outer queue; a producer attempts to dequeue from the inner queue if the limit has been reached and a consumer will push nil items to the queue which frees up the first producer.

I suspect this is too heavy-weight of a tool to use for this and I only chose it because the Queue guarantees order. If a semaphore or other similar primitive is better suited let me know!

### DIFF
--- a/lib/async/queue.rb
+++ b/lib/async/queue.rb
@@ -51,7 +51,7 @@ module Async
 			super()
 			
 			@limit = limit
-			@full = Async::Condition.new
+			@full = Async::Queue.new
 		end
 		
 		attr :limit
@@ -63,7 +63,7 @@ module Async
 		
 		def enqueue item
 			if limited?
-				@full.wait
+				@full.dequeue
 			end
 			
 			super
@@ -72,7 +72,7 @@ module Async
 		def dequeue
 			item = super
 			
-			@full.signal unless @full.empty?
+			@full.enqueue(nil) unless @full.empty?
 			
 			return item
 		end
--- a/spec/async/queue_spec.rb
+++ b/spec/async/queue_spec.rb
@@ -53,4 +53,25 @@ RSpec.describe Async::LimitedQueue do
 		subject.enqueue(10)
 		expect(subject).to be_limited
 	end
+
+	it 'should resume waiting tasks in order' do
+		total_resumed = 0
+		total_dequeued = 0
+		Async do |producer|
+			10.times do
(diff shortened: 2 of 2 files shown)


Bringer128:
Looks like the CI failure is TruffleRuby only and is the same as latest master.

ioquatix:
This looks like a pretty reasonable addition.

There are CI failures.

Perhaps try using semaphore: https://github.com/socketry/async/blob/master/lib/async/semaphore.rb

ioquatix:
Using a queue is super clever BTW, I like it. But yes it might be heavy.

Bringer128:
> There are CI failures.

Oops sorry! My first time using TravisCI.

> Perhaps try using semaphore

I had a look into this and the issue with semaphore is that it uses `pop` on the list of waiting fibers, so it actually ends up reversing the order. The alternative is to create a queue-like condition class e.g.

```ruby
class ConditionQueue < Condition
  def signal(value)
    fiber = @waiting.shift
    fiber.resume(value) if fiber&.alive?
  end
end
```

ioquatix:
I guess we could change semaphore, the order of first in first served probably does make sense. What do you think?

Bringer128:
I don't have a strong opinion about ordering guarantees of a semaphore implementation. Of the implementations I've used Java is FIFO and C# provides no guarantee.

My biggest concern would be with coupling the two types together as it reduces the future flexibility of the semaphore implementation but it's also your library so that's a tradeoff you're better suited to make than me!

I personally would vote for leaving it a Queue unless you have some serious concerns about performance.

ioquatix:
Leaving it as a queue is fine for now, but can you fix the build failures?

Bringer128:
Hopefully that fixes it! If not I'll tackle it soon.

Thanks for being so responsive!

PR #38, opened 2019-03-21
Allow `task.logger=`.

ioquatix:
Related discussions:

- https://github.com/socketry/rubydns/issues/42
- https://github.com/socketry/event/issues/3

The main point is should user be able to change `task.logger`?

Because `Task.logger` is defined as:

```ruby
		def logger
			@logger || @parent&.logger
		end
```

If `parent.logger` is changed, the child logger will automatically use it.

A more efficient approach is:

```ruby
		def logger
			@logger ||= @parent&.logger
		end
```

This will cache the logger, but it won't be correct of the parent logger is changed after the child is created. It's more efficient after first invocation but the semantics are wrong if we also want to support `task.logger=`.

One alternative to `task.logger=` is to make a nested child task. It's a little bit heavy, but it might make things simpler.

e.g.

```ruby
endpoint.accept do |peer, task|
  task.async(logger: RequestLogger.new(task.logger, peer)) do
    # Process request here
  end.wait
end
```

Feedback welcome @bryanp

### DIFF
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -75,6 +75,10 @@ module Async
 			@logger || Event::Console.logger
 		end
 		
+		def logger=
+			@logger = logger
+		end
+		
 		def to_s
 			"<#{self.description} stopped=#{@stopped}>"
 		end
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -72,11 +72,7 @@ module Async
 		end
 		
 		def logger
-			@logger || Event::Console.logger
-		end
-		
-		def logger=
-			@logger = logger
+			@logger ||= Event::Console.logger
 		end
 		
 		def to_s
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -76,11 +76,7 @@ module Async
 		end
 		
 		def logger
-			@logger || @parent&.logger
-		end
-		
-		def logger=
-			@logger = logger
+			@logger ||= @parent&.logger
 		end
 		
 		# @attr ios [Reactor] The reactor the task was created within.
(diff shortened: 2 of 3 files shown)


bryanp:
Between the two, the nested child approach seems to be a better (less complex) path forward. I don't know how much overhead it creates, but maybe it's worth it for a simpler approach? It seems natural for my use case which is built on top of `Async::HTTP::Server`:

```ruby
def call(request)
  connection = Connection.new(request)
  Async::Task.current.async(logger: connection.logger) do
    ...
  end
end
```

Another approach might be to define a "generator" called during `Task` creation that returns the logger instance. It might not work in my particular case, since the logger that I want to be the task logger exists inside of a `Connection` instance, which depends on the request.

ioquatix:
Making a task just for logging isn't a huge overhead, but it's also not trivial.

Adjusting the logger is sort of mutable state which I've tried to avoid. I think we can add mutable logger in the future if it turns out this approach doesn't work. But it's hard to take back mutability once it's in use.

I think the typical use case should be to still use `Async.logger` which dynamically gets the task logger where possible, unless you can access `task.logger`.

Rather than trying to wrap your own logger/request handling, maybe it makes sense for clients and servers to do this implicitly, e.g. when a task is created by `accept_each`, it could automatically create a scoped logger. That being said, it could introduce additional overheads, and ideally logging (especially when being skipped due to filter level) shouldn't introduce a huge overhead.

The question should probably be whether the overheads of using a scoped logger make sense.

- There is a non-zero cost to getting and scoping loggers, even if they wouldn't output anything.
- There is additional surface area for exposing `Reactor#logger` and `Task#logger`.
- The value of scoped logging probably depends more on disparate libraries using them, e.g. trying to associate a SQL query from Sequel to a response being generated by Falcon.

I'd like to know more about what this should look like.

```
3s: <Server> Handling incoming request
3.1s Request@0x1234: Making SQL query
3.2s Request@0x1234: Rendering template
3.3s Request@0x1234: Writing response.
```

PR #52, opened 2019-11-04
Implement before- and after-fiber-creation hooks.

BMorearty:
PR for discussion:

Web apps sometimes store per-request state in a thread-local variable, e.g. using the [request_store gem](https://github.com/steveklabnik/request_store). If a server spawns multiple fibers while handling a request, the fibers won't have access to the thread-local storage of the original fiber.

This PR has a possible solution: give clients the ability to insert one or more hooks before and after fiber creation. Each `before_create` hook is called in the parent fiber and returns a payload. The paired `after_create` hook is called in the new child fiber and is passed in the payload that `before_create` returned.

This enables passing thread-local storage into a task:

    Async::Reactor.hook(
      before_create: -> { Thread.current[:store] }),
      after_create: ->(store) { Thread.current[:store] = store.dup }
    )

(For this to work with `request_store` I could make  PR to add a `replace!(store)` method that replaces the thread-local variable with the store that's passed in.)

There's one potentially large flaw in this approach that is open for discussion: the way I wrote it, the `hook` method can only be called once you're already inside a reactor loop. This makes it impossible for a web app to have an initialization phase at startup that sets up all the hooks, but I think that would be an easier way to use an interface like this. And web apps with Falcon don't own the outer reactor loop, so they can't just write an outer `Async do` followed by a setting up the hooks, as I did in the specs.

### DIFF
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -69,6 +69,13 @@ module Async
 			
 			return NIO::Selector.new
 		end
+
+		# Set up a hook pair to be called back when instantiating a task.
+		# @param before_create [Lambda] will be evaluated before a task's fiber is created. Takes 0 params.
+		# @param after_create [Lambda] will be called in the new fiber and passed the return value of before_create. Takes 1 param.
+		def self.hook(before_create: nil, after_create: nil)
+			Task.current.reactor.hook(before_create: before_create, after_create: after_create)
+		end
 		
 		def initialize(parent = nil, selector: self.class.selector, logger: nil)
 			super(parent)
@@ -273,6 +280,34 @@ module Async
 			warn "#{self.class}\#timeout(...) is deprecated, use #{self.class}\#with_timeout(...) instead."
 			
 			with_timeout(*args, &block)
+		end
+
+		# @return [Array] A list of callbacks to be called before and after creating a fiber.
+		# Each is a Hash with keys :before_create (a lambda to be called in the original
+		# fiber) and :after_create (a lambda to receive the value as a parameter).
+		def hooks
+			@hooks ||= []
+		end
+
+		# Before creating an async task's fiber, call all the before_create hooks and return their values.
+		def call_before_hooks
+			hooks.map { |hook| hook[:before_create]&.call }
+		end
+
+		# In a new async task's fiber, pass the return values of the before_create hooks to the
+		# after_create hooks.
+		def call_after_hooks(payloads)
+			hooks.each_with_index do |hook, index|
+				hook[:after_create]&.call(payloads[index])
+			end
+		end
+
+		# Add a hook pair.
+		def hook(before_create:, after_create:)
+			hooks << {
+				before_create: before_create,
+				after_create: after_create,
+			}
 		end
 	end
 end
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -249,8 +249,10 @@ module Async
 		end
 		
 		def make_fiber(&block)
+			payloads = reactor.call_before_hooks
 			Fiber.new do |*arguments|
 				set!
+				reactor.call_after_hooks(payloads)
 				
 				begin
 					@result = yield(self, *arguments)
--- a/spec/async/reactor_spec.rb
+++ b/spec/async/reactor_spec.rb
@@ -19,6 +19,7 @@
 # THE SOFTWARE.
 
 require 'async'
+require 'async/rspec'
 require 'benchmark/ips'
 
 RSpec.describe Async::Reactor do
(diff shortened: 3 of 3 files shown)


ioquatix:
Tasks can access their parent. Could
You achieve the same result by accessing state on the parent or exposing something to access parent state?

ioquatix:
One more thing to consider is how this impacts life cycle.

e.g.

```
Async do |parent|
	internet = Async::HTTP::Internet.new
	# 1st issue is shared namespace for keys... easy to clobber existing keys.
	# 2nd issue is dealing with life-cycle. Spooky action at a distance, etc.
	parent[:internet] = internet
ensure
	internet&.close
end
```

BMorearty:
Accessing state on the parent seems reasonable too. For my use case it's not much work--just requires replacing `request-store` with a new gem that accesses state in `Async::Task.current`, which is simple. But it does have a disadvantage in that all thread-local variables (that one wants to share across fibers) must have their storage moved to the async task. If you're using any open source gems with thread-local storage you'd have to modify them all. Using the hook approach, the changes to external gems seem likely to be smaller. 

I agree with your points about the life cycle and clobbering existing keys. The reason I used `dup` in the sample code was so the child gets a fork of the parent's state instead of referencing the original later on, when it might have changed.

ioquatix:
The entire thread-local/fiber-local issue and global state is kind of a mess. Code expecting it to be thread local, but it's not, etc.

Looking forward, these issues might be forced by removing Thread, thread locals, etc. That could happen with hypothetical guilds. Fiber local might be the default.

Best to find a theoretical model for how this should work then implement it in the code. LISP has dynamic variable scope... which is similar to global but scoped on call chains. Its model makes more sense when the stack represents the flow of code w.r.t. request handling (assuming each request is handled by invoking a method which wraps the dynamic state). This is pretty natural for fibers, but might present some challenges to the implementation if fibers are considered unique stacks.

It lets you do something like:

```ruby
def request
	$per_request_state = {...}
	middleware.call
end

# ... somewhere in middleware
$per_request_state[...]
```

$per_request_state effectively looks up the call chain to find the inner most definition. It's called dynamic scope.

I know we want to find a short term solution and there is definitely a gap in the interface here, but I don't want to force users to deal with thread locals since it might break in the future (or it's already broken, depending on your opinion :)

Right now, the only solution is to actually pass the state along and that's kind of a good default since it forces people to think about state and lifetimes. I know this isn't a good answer for legacy code...

There is only one example of code which behaves this way: `Task#logger` which scopes logging to a task. The behaviour is very specific to avoid performance issues - i.e. you can't change the logger after the task is created.

To me, `Task#[]` is kind of a hack. It's not easy to use correctly (clobbering), it might be hard to implement efficiently `O(N)` for N deep task tree, so the motivation to implement it is kind of low.

I don't think that's the right answer so I need to play around with it a bit more.

ioquatix:
> If you're using any open source gems with thread-local storage you'd have to modify them all. Using the hook approach, the changes to external gems seem likely to be smaller.

This makes a lot of sense to me.

ioquatix:
It might be good to have a chat about this... but I just wondered how the following would work:

```
# parent task sets some fiber local state
Thread.current[:foo] = "bar"

# makes child fiber
Thread.current[:foo]

# Parent is executing
Thread.current[:foo] = "baz"

# Child is executing
Thread.current[:foo] # what is value? "bar" or "baz"
```

BMorearty:
I'm not sure what question you are asking with your code sample, but my conclusion was that normally you'd want the child fiber to have a copy of the parent's thread state, not a reference to it. E.g., if the parent sets a thread-local variable to an object: `Thread.current[:foo] = { bar: "baz" }` and the child accesses `Thread.current[:foo]`, you'd want it to be accessing a copy of `{ bar: "baz" }`.

Here's why:

```ruby
# Parent sets request context:
Thread.current[:store] = { user_id: 123 ]

# Makes child fiber which also needs access to the request store:
Thread.current[:store][:user_id] # <= 123 if we copied the thread-local storage down

# Parent finishes processing the request and cleans up local storage:
Thread.current[:store].clear

# Child resumes execution and still needs to access `user_id`;
Thread.current[:store][:user_id] # <= You want 123, not nil
```

I can't think of a use case for the child to be able to modify the fiber state of the parent. That's why I'm fine with `dup`ing the state into the child. But the nice thing about the hooks approach is its flexibility. It's up to the client to decide how to handle the data and whether to `dup` it.

ioquatix:
You got the general idea of the code example, it's to do with by reference vs by value.

The problem with global hooks is it's a general policy at the reactor level, not at the code level, while it might make some operations more compatible, it's possible to cause failures too (i.e. some code might expect by value, other code might expect by reference, and thus mixing together the two libraries fail).

In order to do this kind of thing, I don't believe it should be determined by user policy, there should be one logical way which solves this problem.

ioquatix:
I'm going to close this issue.

It introduces global state which I've tried to avoid in the design of async.

The solution is to explicitly share state where required.

The alternative is to use real thread locals, e.g. https://github.com/socketry/thread-local is one option.

If you use real thread locals you don't need to duplicate the state as you go between tasks. The disadvantage with thread locals is they are effectively global, where as the PR here propagates data based on the task hierarchy.

However, if you want that, my suggestion is you do it explicitly.

ioquatix:
Thanks for your effort :)

PR #54, opened 2020-01-09
Left pad Node#inspect hex digits with zeros.

havenwood:
This tries to adjust `Node` #inspect to more closely resemble the vanilla format by adding left padded zeros. I don't know if less padding would be preferred, but it seems nice to left pad a bit so printed hierarchies align and such. Maybe eight hex digits?

For comparison, this is before.

```
#<Async::Node:0x8c>
	#<Async::Node:0x3c>
		#<Async::Node:0x50>
			#<Async::Node:0x64>
	#<Async::Node:0x78>
```

And after.

```
#<Async::Node:0x0000000000000118>
	#<Async::Node:0x0000000000000078>
		#<Async::Node:0x00000000000000a0>
			#<Async::Node:0x00000000000000c8>
	#<Async::Node:0x00000000000000f0>
```

Or with 8 rather than 16 for padding.

```
#<Async::Node:0x00000118>
	#<Async::Node:0x00000078>
		#<Async::Node:0x000000a0>
			#<Async::Node:0x000000c8>
	#<Async::Node:0x000000f0>
```

Or even just 6.

```
#<Async::Node:0x000118>
	#<Async::Node:0x000078>
		#<Async::Node:0x0000a0>
			#<Async::Node:0x0000c8>
	#<Async::Node:0x0000f0>
```

I also added a left shift of the object ID, like has been done with inspect in recent Rubies, but I'm unsure if that's still worth doing now that Ruby 2.7 is doing IDs differently.

Ruby 2.6:
```ruby
o = Object.new
#=> #<Object:0x0000000114858260>

format('%#018x', object_id << 1)
#=> "0x0000000114858260"
```

Ruby 2.7
```ruby
o = Object.new
#=> #<Object:0x000000011428f590>

format('%#018x', object_id << 1)
"0x0000000000000190"
```

I don't know if you want to left pad or left shift but I got to making the PR before I realized object IDs changed in Ruby 2.7. Are sixteen hex digits too many to pad? I personally prefer padding 8 to avoid the varied length.

Thanks for the great library in any case!


ioquatix:
Please follow width of Ruby 2.7 and add a spec for the formatting. Thanks.

havenwood:
Updated, added tests and rebased.

olleolleolle:
@ioquatix This looks ready to go in!

ioquatix:
Thanks. I'm going to merge this. I might try it out before releasing it or I might modify it, but having some extra specs is great. The main thing is that object ids and memory addresses are different. I feel like formatting rules might need to reflect that, since typically object addresses are formatted with enough digits because we know the minimum and maximum bounds. For Ruby object ids, we could even know that and it won't be 2^64 for sure.

PR #55, opened 2020-01-09
Add a Node#hierarchy method for flexible accounting.

havenwood:
This PR adds a Node#hierarchy for more flexibility with outputting hierarchy. I switched over Node#print_hierarchy to use the new method internally.

My hope was to make Node hierarchy discoverable via a mechanism other than printing.

For an example, here's a typical #print_hierarchy output.

```
#<Async::Node:0x000118>
	#<Async::Node:0x000078>
		#<Async::Node:0x0000a0>
			#<Async::Node:0x0000c8>
	#<Async::Node:0x0000f0>
```

And the corresponding current #hierarchy return value.

```ruby
[{:level=>0, :node=>#<Async::Node:0x000118>},
 {:level=>1, :node=>#<Async::Node:0x000078>},
 {:level=>2, :node=>#<Async::Node:0x0000a0>},
 {:level=>3, :node=>#<Async::Node:0x0000c8>},
 {:level=>1, :node=>#<Async::Node:0x0000f0>}]
```

### DIFF
--- a/lib/async/node.rb
+++ b/lib/async/node.rb
@@ -23,6 +23,12 @@ require 'set'
 module Async
 	# Represents a node in a tree, used for nested {Task} instances.
 	class Node
+		HIERARCHY_FORMAT = ->(node, level) { {level: level, node: node} }
+		private_constant :HIERARCHY_FORMAT
+		
+		HIERARCHY_PRINT_FORMAT = ->(node, level) { "#{"\t" * level}#{node}" }
+		private_constant :HIERARCHY_PRINT_FORMAT
+		
 		# Create a new node in the tree.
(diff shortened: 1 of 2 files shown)


ioquatix:
Do you mind explaining the expected use case? Because we have traverse which is very close to the implementation.

havenwood:
> Do you mind explaining the expected use case? Because we have traverse which is very close to the implementation.

If I recall, I was using #traverse to return a collection of nodes, like `.traverse.map { |node, value|` would but got a `LocalJumpError`. In retrospect, I think I should have proposed an `return enum_for(__method__) { size } unless block_given?` form for #traverse rather than a new method.

PR #56, opened 2020-01-11
Scheduler implementation

ioquatix:
This is a prototype of the scheduler which uses the interface provided by https://github.com/ruby/ruby/pull/3032

At the present time, the scheduler is an optional set of hooks for implementing non-blocking IO. However, in the future, the scheduler might form the core of the reactor, replacing the selector entirely.

### DIFF
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -23,6 +23,7 @@
 require_relative 'logger'
 require_relative 'task'
 require_relative 'wrapper'
+require_relative 'scheduler'
 
 require 'nio'
 require 'timers'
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -28,6 +28,7 @@ matrix:
       env: JRUBY_OPTS="--debug -X+O"
     - rvm: truffleruby
     - rvm: ruby-head
+    - rvm: git@github.com:ioquatix/ruby#thread-selector
```
ioquatix: How can I make this work?

havenwood: I _think_ it'd be:

```yaml
- rvm: ruby-head-scheduler --url "https://github.com/ioquatix/ruby.git" --branch selector
```
P.S. - I meant to push a commit to my fork of your brach to test CI, not your branch itself. Reverted. I'm sorry!

havenwood: Nope, I'm wrong. It doesn't work since Travis is appending a `--binary` flag. https://travis-ci.org/havenwood/async/jobs/635745650

ioquatix: Since this is now merged into master, this is no longer a problem *for this PR*. Thanks for your help.

PR #58, opened 2020-02-10
Make Kernel#Sync public

sailor:
Hello,

I'm not sure if it was intended to be public, but i would love to use `Sync` for some use cases. The one i encountered is that i need the value returned by a call to some `async-redis` function.

I often ends up doing this:

```ruby
value = Async { redis.llen 'list' }.wait
```

which is exactly what is provided by the `Kernel#Sync` method. Do you think that exposing this method could make sense?

Thank's for this awesome gem!

### DIFF
--- a/lib/async.rb
+++ b/lib/async.rb
@@ -25,6 +25,7 @@ require_relative "async/logger"
 require_relative "async/reactor"
 
 require_relative "kernel/async"
+require_relative "kernel/sync"
 
 module Async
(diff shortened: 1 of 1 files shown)


ioquatix:
Okay. But maybe we also need to add documentation.

ioquatix:
Until we make an official release of this (which I might need to spend some more time thinking about), 

```
require 'kernel/sync`
```

will work for the foreseeable future.

PR #59, opened 2020-02-22
Sync method should pass the task back.

BMorearty:
I'd like to be able to call `Sync do |task|` and have it pass the task to my callback so I can access the task without another call to `Async::Task.current`.

Added a test for this.


ioquatix:
That makes total sense. Thanks.

PR #71, opened 2020-06-10
Adds a method to DRY timer code a little bit

wyhaines:
Timers are simple to create with Async, but they require repeating an incantation of boilerplate for every timer that an app creates.

This method wraps that incantation so that instead of this:

```ruby
async do |timer_task|
  loop do
    timer_task.sleep DURATION # some number of seconds to wait between runs of the timer
   # Do Stuff Here
  end
end
```

One can do this:

```ruby
task.add_timer(DURATION) do
  # Do Stuff Here
end
```

It's a little cleaner to look at, and it's a little more clear when scanning the code what is happening.

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -178,6 +178,17 @@ module Async
 			end
 		end
 		
+		# Syntactic sugar to create a timer.
+		def add_timer(interval: 0, periodic: true)
+			async do |subtask|
+				loop do
+					subtask.sleep(interval)
+					yield
+					break unless periodic
+				end
+			end
+		end
+
 		# Lookup the {Task} for the current fiber. Raise `RuntimeError` if none is available.
 		# @return [Async::Task]
 		# @raise [RuntimeError] if task was not {set!} for the current fiber.
--- a/spec/async/task_spec.rb
+++ b/spec/async/task_spec.rb
@@ -315,6 +315,47 @@ RSpec.describe Async::Task do
 			expect(state).to be == :finished
 		end
 	end
+
+	describe '#add_timer' do
+		let(:duration) {0.01}
+
+		it "will fire a single use timer after the requested duration" do
+			state = nil
+
+			time = Async::Clock.measure do
+
+				reactor.async do |task|
+					task.add_timer(interval: 2, periodic: false) { state = :finished }
+					task.sleep(duration * 2)
+				end
+
+				reactor.run
+			end
+
+			expect(state).to be == :finished
+			expect(time).to be >= duration
+		end
+
+		it "will fire a periodic use timer after the requested duration" do
+			state = 0
+
+			time = Async::Clock.measure do
+
+				reactor.async do |task|
+					task.add_timer(interval: duration) do
+						state += 1
+						break if state > 1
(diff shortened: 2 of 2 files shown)


ioquatix:
Can you show me the use cases for this functionality, thanks!

wyhaines:
Sure!

```
task.add_timer(interval: 1) { Analogger.update_now }
task.add_timer(interval: @config['interval']) { write_queue }
task.add_timer(interval: @config['syncinterval']) { flush_queue }
```

vs.

```
async do |update_task|
  loop do
    update_task.sleep 1
    Analogger.update
  end
end

async do |write_task|
  loop do
    write_task.sleep @config['interval']
    write_queue
  end
end

async do |sync_task|
  loop do
    sync_task,sleep @config['syncinterval]
    flush_queue
  end
end
```

Even if writing it with the most compact Ruby style, the difference is:

```
task.add_timer(interval: 1) { Analogger.update_now }
```

vs.

```
async {|_| loop { _.sleep(1); Analogger.update_now }}}
```

Which is fewer characters, but is a lot more obtuse.

ioquatix:
I understand your use case.

The next point is naming.

Can you consider the existing methods:

```
Reactor#after(duration) do ... end
```

and

```
Reactor#every(duration) do ... end
```

These methods do not create asynchronous task and work directly in the event loop. My feeling was this was too risky for normal use case, users might try to implement health check for example, which would break if they needed to do I/O or some other non-blocking task.

As such these methods are deprecated and are not used in any code that I'm aware of. So we can repurpose them.

In addition, we can introduce `Task#after` and `Task#every` which *could* easily introduce a new asynchronous context like your proposal.

So, can you consider these names and let me know what you think and how your design can fit into the existing design?

ioquatix:
(@bryanmacfarlane it is confusing that this PR is marked as failing, but the failing test is marked as `continue-on-error`, not sure if there is some way to deal with this in a more practical/useful way)

wyhaines:
Thanks. I will take a look at those methods today and refactor the PR to use them. The intent expressed by the naming seems to me to fit well. I like it.

ioquatix:
Great, I look forward to your updated PR. For now, I'm going to bed.

ioquatix:
By the way, we might not need to change `Reactor#after` and `Reactor#every` methods. My point was supposed to be, it exists, it's deprecated (ish), maybe we can change it, but introducing `Task` related methods also makes sense for me and is the main entry point for users.

ioquatix:
I'm still keen to see an update to this PR :)

ioquatix:
Any update?

ioquatix:
I will implement this myself.

ioquatix:
I played around with the implementation and decided against introducing it.

- You cannot use `break` correctly for some reason (maybe ruby bug).
- There are at least four permutations, maybe more: do it once after a timeout, do it now and once after a timeout, do it now and repeat after a timeout, wait for the timeout and do it, then repeat, etc.
- With Ruby 3, bare `sleep` will be hooked by async, so it's less cumbersome.
- We should encourage users to write event-loop agnostic programs where possible.

PR #72, opened 2020-08-17
Outline of how to support mutex.

ioquatix:


lib/async/reactor.rb
```diff
@@ -148,6 +148,22 @@ def << fiber
 			@ready << fiber
 		end
 		
+		# Resume the given fiber on the next iteration of the reactor. If urgent, try to wake up the reactor if it is currently waiting for events.
+		# @parameter fiber [Object] Any object that responds to `#resume`.
+		# @reentrant Thread safe. Fiber will be invoked in the context of the scheduler's thread.
```
olleolleolle, line 153: Nice annotation!

PR #75, opened 2020-10-01
Allow setting custom Async::Queue queue

bruno-:
### Problem

I want to use `Async::Queue` with a custom queue class.

## Types of Changes

- [ ] Bug fix.
- [x] New feature.

## Testing


ioquatix:
This seems okay. However, can you rename `queue:` argument to `items:` or can we think of a better ivar name?

I always feel weird when we have something like `Queue.new(queue: ...)` since the names are the same.

Also sorry I missed this.

bruno-:
I agree `Queue.new(queue: ...)` looks weird.
We could go with `items` or `elements` but usage like `Queue.new(items: LifoQueue.new)` is also not ideal. "items" and "lifo queue" are not at the same level of abstraction as "lifo queue" contains "items". `internal_data_structure` would be perfect, but that's too long.

I spent some time thinking about this and here are my top picks:

1. `state`
2. `store`
3. `values`
4. `structure`, `stream`, `sequence`, `container`, `construct`, `collection`, `buffer`, `list`, `nodes`, `internal`, `data` (all bad)

What are your thoughts? Let me know and I'll make a change. Still want to go with `items`?

bruno-:
Another consideration: `Async::Semaphore` and `Async::Notification` could also accept the same argument. It could be a naming pattern, similar to like `@parent` is a parent task.

In fact, I wanted to submit the same PR for `Async::Semaphore` but then my "needs" grew and I created [async-limiter](https://github.com/bruno-/async-limiter). What do you think about moving `async-limiter` to socketry github organization?

ioquatix:
Yes, feel free to move that repo here, thanks! Send it to me first.

Regarding the best name, yes, `items` seems wrong in a general sense.

The problem is, a queue always works in order. We should perhaps consider a generic "queue", but then it should not use the same abstraction/method/class name IMHO. I'll think about it over the weekend.

ioquatix:
I took a look at your gem, and I think it makes more sense than `Async::Queue` or `Async::Semaphore`.

bruno-:
> Send it to me first.

I have just transferred `async-limiter` ownership to your username `ioquatix`.

bruno-:
> The problem is, a queue always works in order. We should perhaps consider a generic "queue", but then it should not use the same abstraction/method/class name IMHO.

Hmm.. so I had this use case where I wanted to use `Async::Queue` backed by a "priority queue" data structure. "Priority queue" doesn't always work in order. But yea, that's an exception and maybe "priority queue" should really be named "priority heap" or something.

> I'll think about it over the weekend.

Thanks. I also understand you're working on Ruby 3 scheduler so I'm ok if you want to postpone this one.

spec/async/queue_spec.rb
```diff
@@ -80,6 +80,44 @@
 			expect(count).to be == repeats
 		end
 	end
+
+	context 'with a custom queue' do
+		class LIFOQueue
```
ioquatix, line 85: Unfortunately RSpec does not introduce scope, so this is global.

bruno-: I just fixed this.

But I think the main problem with this PR was that it got stuck with naming the internal queue...

ioquatix: Yes, will revisit this.

ioquatix (approved):

PR #77, opened 2020-12-10
Async::Task: don't forward #timeout to reactor

paddor:
As far as I can tell, `Async::Reactor` does not have a `#timeout` method.

## Types of Changes

- [X] Bug fix.
- [ ] New feature.

## Testing

- [ ] I added new tests for my changes.
- [X] I ran all the tests locally.


ioquatix:
Thanks, you are correct, it was deprecated and removed.

Issue #86, opened 2020-12-30
wrong number of arguments with Scheduler


ankane:
Hey @ioquatix, thanks for this neat project! I was trying out the new scheduler with [this example](https://rubyreferences.github.io/rubychanges/3.0.html#non-blocking-fiber-and-scheduler) and ran into a few errors.

```ruby
require "async"
require "net/http"

reactor = Async::Reactor.new
start = Time.now

Thread.new do 
  Fiber.set_scheduler(reactor.scheduler)

  %w[2.6 2.7 3.0].each do |version|
    Fiber.schedule do
      t = Time.now
      Net::HTTP.get('rubyreferences.github.io', "/rubychanges/#{version}.html")
      puts '%s: finished in %.3f' % [version, Time.now - t]
    end
  end
end.join
```

Stack trace

```text
/Users/user/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/async-1.28.2/lib/async/task.rb:75:in `initialize': wrong number of arguments (given 0, expected 1..2) (ArgumentError)
	from /Users/user/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/async-1.28.2/lib/async/scheduler.rb:105:in `new'
	from /Users/user/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/async-1.28.2/lib/async/scheduler.rb:105:in `fiber'
	from go.rb:14:in `schedule'
	from go.rb:14:in `block (2 levels) in <main>'
	from go.rb:13:in `each'
	from go.rb:13:in `block in <main>'
```

It looks like `@reactor` needs to be passed to `Task.new`. After adding that, ran into `undefined method 'resume' for #<Async::Task>`.

ioquatix:
You need to use the `Async` construct as outlined in the documentation.

ankane:
My bad, from #56, I thought it was an implementation of `Fiber::SchedulerInterface`.

ioquatix:
It is.

PR #87, opened 2021-01-02
Fix Async::Scheduler#fiber hook

ysbaddaden:
The current implementation of `Async::Scheduler#fiber(&block)` causes `Fiber.schedule {}` to always fail. For example:

```ruby
require "async"

Async do
  Fiber.schedule { puts "hello" }
end
```

```
  0.0s    error: Async::Task [oid=0x3c] [pid=51] [2021-01-02 10:36:33 +0000]
               |   ArgumentError: wrong number of arguments (given 0, expected 1..2)
               |   → vendor/gems/3.0.0/gems/async-1.28.2/lib/async/task.rb:75 in `initialize'
               |     vendor/gems/3.0.0/gems/async-1.28.2/lib/async/scheduler.rb:105 in `new'
               |     vendor/gems/3.0.0/gems/async-1.28.2/lib/async/scheduler.rb:105 in `fiber'
               |     bug.rb:4 in `schedule'
               |     bug.rb:4 in `block in <main>'
               |     vendor/gems/3.0.0/gems/async-1.28.2/lib/async/task.rb:265 in `block in make_fiber'
```

## Types of Changes

- [x] Bug fix.
- [ ] New feature.

This tiny patch fixes the calls to `Async::Task` and allows the usage of `Fiber.schedule` to create non blocking fibers.

## Testing

I tried to write a test, but I'm not knowledgeable enough with how the spec are written.

### DIFF
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -102,9 +102,9 @@ module Async
 		end
 		
 		def fiber(&block)
-			task = Task.new(&block)
+			task = Task.new(@reactor, &block)
 			
-			task.resume
+			task.run
 			
 			return task.fiber
 		end


ysbaddaden:
Fixes the crash reported in #86, but doesn't make it work, because we must wrap the app inside an `Async` block to _start_ the reactor.

ioquatix:
Thanks!

PR #89, opened 2021-01-11
Async::Task#wait_all

bryanp:
Provides a way to wait on all children, _including_ children and their tasks. This came out of a conversation I had with @ioquatix last week and is the approach I ended up moving forward with for my own use-case.

## Types of Changes

- [ ] Bug fix.
- [x] New feature.

## Testing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,4 @@ root = true
 [*]
 indent_style = tab
 indent_size = 2
+trim_trailing_whitespace = false
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -156,6 +156,16 @@ module Async
 		# Deprecated.
 		alias result wait
 		# Soon to become attr :result
+
+		# Wait on all children to complete, including subtasks.
+		# @return [Object] the final expression/result of the task's block
+		def wait_all(task = self)
+			task.children&.each do |child|
+				wait_all(child)
+			end
+
+			task.wait
+		end
 		
 		# Stop the task and all of its children.
 		# @return [void]
--- a/spec/async/task_spec.rb
+++ b/spec/async/task_spec.rb
@@ -468,6 +468,52 @@ RSpec.describe Async::Task do
 			expect(innocent_task).to be_finished
 		end
 	end
+
+	describe '#wait_all' do
+		it "will wait on all subtasks to complete" do
+			result = nil
+
+			reactor.async do |task|
+				wait_task = task.async do |subtask1|
+					subtask1.async do |subtask2|
+						subtask2.async do |subtask3|
+							subtask3.sleep(0.25)
+
+							result = :subtask3
+						end
+					end
+				end
+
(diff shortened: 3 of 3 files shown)


chuckremes:
I'd like to see this feature added too. Useful.

ioquatix:
It's useful but it's also risky because some interfaces will create tasks for persistent connections that will outlive your own tasks and if you try to wait on them you will be waiting forever.

bryanp:
Coming back to this, I think it would help for `async` to define what kind of library it is. Right now the decision to not include this feature feels arbitrary. There are lots of ways for users to trip over themselves—why exclude this particular one? IMO defining the types of problems the library solves vs what the user is left to solve would help inform these discussions.

trevorturk:
I've been thinking about this (shower thoughts) and I don't know if this is helpful, but I've been thinking of `Async::Barrier` as less of a "barrier" and more of a "bucket" and I wonder if that mindset is valid, and, if so, might be helpful when considering this. When I'm using a barrier, I'm thinking "let's toss this in that bucket, and this over here, and maybe this, too" and then calling `wait` I believe essentially means `wait_all` within that "bucket" context. 

```ruby
bucket = Async::Barrier.new
barrier.async { some_expensive_work }
some_other_method
something_over_here
do_this_in_the_bucket_too(bucket)
bucket.wait
```

I'm still trying to wrap my head around all of this async and/or concurrency stuff (I'm also learning Swift and SwiftUI now) but I wonder if this kind of framing is helpful at all -- tl;dr calling `Async::Barrier.wait` is essentially `wait_all` for that group of tasks. 

I was also wondering if it would be useful to be able to "tag" async tasks and then `wait_all_for_tag(tag)` but I'm not sure if that's any different from the barrier concept, really...

Anyway, just some food for thought on this topic! I hope I'm not too far off base! :)

ioquatix:
@trevorturk that's pretty accurate.

> Right now the decision to not include this feature feels arbitrary

@bryanp it's not arbitrary, it's because the risk of `wait_all` stalling indefinitely/causing a deadlock is quite high AND might not be obvious in testing vs in production.

Suppose you have an `Async::HTTP::Internet` instance. A hypothetical `wait_all` implementation might end up waiting on an internal task created by HTTP/2 for persistent connections. It's better to use `Async::Barrier` and be explicit about things, or wait on a parent task which explicitly waits on its children (or not) and so on.

Even if `wait_all` existed, I would never recommend people use it, which is why I hesitate to introduce it.

ioquatix:
Thanks everyone for the discussion. For the reasons given above, I'm going to hold off on this PR for now.

samuel-williams-shopify:
> it's because the risk of wait_all stalling indefinitely/causing a deadlock is quite high AND might not be obvious in testing vs in production.

I've proposed https://github.com/socketry/async/pull/441 which explicitly ignores transient tasks, so that issue no longer applies.

PR #94, opened 2021-01-20
Expose task finished condition so that it can be created optimistically to avoid race on #wait.

ioquatix:
## Types of Changes

- [x] Bug fix.
- [x] New feature.

## Testing

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -72,14 +72,14 @@ module Async
 		# Create a new task.
 		# @param reactor [Async::Reactor] the reactor this task will run within.
 		# @param parent [Async::Task] the parent task.
-		def initialize(reactor, parent = Task.current?, logger: nil, **options, &block)
+		def initialize(reactor, parent = Task.current?, logger: nil, finished: nil, **options, &block)
 			super(parent || reactor, **options)
 			
 			@reactor = reactor
 			
 			@status = :initialized
 			@result = nil
-			@finished = nil
+			@finished = finished
 			
 			@logger = logger
 			
--- a/lib/kernel/sync.rb
+++ b/lib/kernel/sync.rb
@@ -28,7 +28,10 @@ module Kernel
 		if task = ::Async::Task.current?
 			yield task
 		else
-			::Async::Reactor.run(&block).wait
+			::Async::Reactor.run(
+				finished: ::Async::Condition.new,
+				&block
+			).wait
 		end
 	end
 end
--- a/spec/kernel/sync_spec.rb
+++ b/spec/kernel/sync_spec.rb
@@ -43,12 +43,22 @@ RSpec.describe Kernel do
 				result = Sync do |sync_task|
 					expect(Async::Task.current).to be task
 					expect(sync_task).to be task
-
+					
 					next value
 				end
 				
 				expect(result).to be == value
 			end
+		end
+		
+		it "can propagate error without logging them" do
+			expect(Console.logger).to_not receive(:error)
+			
+			expect do
+				Sync do
+					raise StandardError, "brain not provided"
+				end
+			end.to raise_exception(StandardError, /brain/)
 		end
 	end
 end


igor-makarov:
@ioquatix that helps! The CocoaPods tests now don't output log noise.

However, don't you think this affects all tasks, and not just `Sync`?

igor-makarov:
Example:
```
expect(Console.logger).not_to receive(:error)

task = Async do |task|
	task.async do 
		raise StandardError, "brain not provided"
	end.wait
end

expect do 
	task.wait
end.to raise_exception(StandardError, /brain/)
```

ioquatix:
The logging does not affect behaviour.

igor-makarov:
@ioquatix however, it does affect diagnostics in production - if errors are printed to production log as false positives, this will be bad for using the library.

Silencing errors in production isn't a good solution either.

igor-makarov:
@ioquatix perhaps `@finished` should be always initialized in the initializer?

ioquatix:
@igor-makarov if you do that, you end up silencing tasks that fails, and if no one calls wait, then the error is lost. Maybe it's okay, but in my experience, it causes developers to miss errors in their code.

ioquatix:
Because the semantics of `Sync` should always be wait, I'm okay with this PR. But for the general case, I'm not sure it's good.

jasl:
> @igor-makarov if you do that, you end up silencing tasks that fails, and if no one calls wait, then the error is lost. Maybe it's okay, but in my experience, it causes developers to miss errors in their code.

I think in CocoaPods, we finaly rescue all errors in main context, but the problem is when task raise error, the logger will first output it, the pass the error to main context.

Maybe redirect async's logger to file (or somewhere) is a better idea?

igor-makarov:
I've added eager `@finished` init and ran all tests - nothing failed. Perhaps it's a good idea to move forward?

jasl:
> I've added eager `@finished` init and ran all tests - nothing failed. Perhaps it's a good idea to move forward?

cool!

igor-makarov:
@ioquatix let me see if I understand it correctly:

* Calling `#wait` on a task propagates the exception.
* Not calling `#wait` on a task at all is bad, as the exception will not be propagated. This is why the log print happens.
* If a task finishes before`#wait` is called, log print will occur because it doesn't know whether `#wait` get called.
* `#wait` can only be called if the task yields.

#### Therefore:
An exception raised in an unyielding task will get printed even if `#wait` is actually called.

ioquatix:
> An exception raised in an unyielding task will get printed even if #wait is actually called.

Yes, but the key point is, if you handle it, even just by `rescue => error; error`, you won't get anything printed.

I don't feel that this is an unreasonable default, given the problems caused by missing errors.

The only case where I think doing this by default is when executing `Sync` which has the semantics of:

- If there is no reactor, create it and run it until completion or error, or
- If there is a reactor, run the code until completion or error.

The `Sync` semantics are designed for very high level entry points into code which is asynchronous but should look synchronous from the outside world no matter what (and efficiently). Actually the model for `Sync` is that it should be semantically the same as:

```
Async do
  # ...
end.wait
```

Whether an error is printed as a side effect is not specified, but I agree we could argue that it shouldn't because we always know `Sync` will call wait.

bruno-:
What do you think about defaulting `finished` to `Async::Condition.new` in this case:

```ruby
task.async(finished: true) do
  # work
end.wait
```

ioquatix:
I think the only valid use case is for `Sync`. I don't know if we should even document how that works (i.e. the option to `Task.new` should not be a public interface).

ioquatix:
I thought about `finished: true` but I'm not sure there is a valid use case for it and it sits on the critical code path without any value in 99.99999999% of cases.

ioquatix:
Okay, it's merged. `Sync` will never output exception to logger.

PR #95, opened 2021-01-22
Add Async::Semaphore#sync

bruno-:
This PR adds `Async::Semaphore#sync`. The feature is equivalent to
`Async::Semaphore#acquire` with a block.

It adds a nice symmetry to `Semaphore#async`, and it's similar to`Kernel#Sync`
and `#sync` from `async-await`. The motivation is API aesthetics.

If this PR is accepted, we could simplify `#acquire` by leaving it for
manual use only (acquire + release), no block argument.

## Types of Changes

- [ ] Bug fix.
- [x] New feature.

## Testing

### DIFF
--- a/lib/async/semaphore.rb
+++ b/lib/async/semaphore.rb
@@ -65,6 +65,16 @@ module Async
 			end
 		end
 		
+		def sync(*arguments)
+			acquire
+			
+			begin
+				yield Task.current, *arguments
+			ensure
+				release
+			end
+		end
+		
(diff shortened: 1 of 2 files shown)


bruno-:
After reading the discussion in #94 and getting a better understanding of `sync` I now think we shouldn't be adding methods willy-nilly or just because they look nice

> The Sync semantics are designed for very high level entry points into code which is asynchronous but should look synchronous from the outside world no matter what (and efficiently)

PR #97, opened 2021-01-27
Prefer weak map for storing wrappers.

ioquatix:
## Types of Changes

- Bug fix.

## Testing

- [ ] I ran all the tests locally.

### DIFF
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -102,7 +102,7 @@ module Async
 			fiber = Fiber.current
 			
 			if timeout
-				timer = self.after(timeout) do
+				timer = @timers.after(timeout) do
 					if fiber.alive?
 						fiber.resume(false)
 					end
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -89,7 +89,7 @@ module Async
 		end
 		
 		def kernel_sleep(duration)
-			@reactor.sleep(duration)
+			self.block(nil, duration)
 		end
 		
 		def block(blocker, timeout)
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -20,10 +20,6 @@
 
 require_relative 'clock'
 
-class IO
-	attr :async_wrapper
-end
-
 module Async
 	class Scheduler
 		if Fiber.respond_to?(:set_scheduler)
@@ -51,7 +47,7 @@ module Async
 		end
 		
 		private def from_io(io)
-			io.async_wrapper ||= Wrapper.new(io, @reactor)
+			Wrapper.new(io, @reactor)
 		end
 		
 		def io_wait(io, events, timeout = nil)
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -19,7 +19,10 @@
 # THE SOFTWARE.
 
 require_relative 'clock'
-require 'objspace'
+
+class IO
+	attr :async_wrapper
+end
 
 module Async
 	class Scheduler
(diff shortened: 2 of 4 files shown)


bruno-:
Hey,
thank you for fixing the problem so fast!

I wanted to learn about weak maps so I played with this PR. I can get this modified spec to pass consistently.
Do you think this is a problem?

```ruby
it "can clean up unused wrappers" do
	input, output = IO.pipe
	
	5.times do # write + read multiple times
		reader = reactor.async do
			input.read(1)
		end
		
		output.write(".")
		
		reader.wait
		10.times do
			GC.start
			sleep(0.001)
		end
		# wrapper is GCed
		expect(reactor.scheduler.wrappers.size).to be == 0

		# But inputs are still open, referenced and will be used again!
		expect(input).to be_an IO
		expect(input).not_to be_closed
		expect(output).to be_an IO
		expect(output).not_to be_closed
	end
end
```

ioquatix:
I think that behaviour is okay?

bruno-:
> I think that behaviour is okay?

Hmm.. ok. Based on some reading about weak map I expected the weak map reference to live as long as the original object reference lives.

I expected the weak map to keep Wrapper objects while the underlying IOs are referenced and kept in memory. In the case of the above spec I expected the scheduler to always contain wrappers because `input` and `output` IOs are referenced and present in memory the whole time.

I may be getting something wrong, or maybe this is a ruby WeakMap problem. If you think this is ok, then 👍 .

bruno-:
Ok, I think I now better understand how `ObjectSpace::WeakMap` works. Here are the lines from ruby source that provided some insight (that's method definition for `ObjectSpace::WeakMap#[]=`):

https://github.com/ruby/ruby/blob/1a3343cfdcb765b94bf63375d45618130d6b92c9/gc.c#L11465-L11470

So, weakmap defines a finalizer for both map key AND value. If either map key or value are dereferenced, the pair is removed from the map when the *next GC runs*.

Related to this PR and `Async::Scheduler`: `Async::Wrapper` instances are not referenced anywhere outside `@wrappers` weak map and they will **all be garbage collected on the next first GC**. The only exception is if `Async::Wrapper` is actively being used in `Async::Scheduler#io_wait` when the GC runs.

Here's the updated spec where it's shown `@wrappers` gets cleared after just one GC:

```ruby
it "cleans all wrappers on the first GC" do
	input, output = IO.pipe
	
	5.times do
		reader = reactor.async do
			input.read(1)
		end
		
		output.write(".")
		
		reader.wait
		sleep(0.001)
		GC.start
		# wrapper is cleared after just one GC!
		expect(reactor.scheduler.wrappers.size).to be == 0
	end
end
```

ioquatix:
You are correct. I did not anticipate this behaviour. I've implemented things similar to `WeakMap` before as an attached cache, and normally apply the "non-retained" logic to the key.

Because of that, I think the IO should probably have the wrapper attached directly to it. It seems like the only way.

ioquatix:
@bruno- what do you think about this updated implementation?

ioquatix:
The problem with this approach is that if you try to use the IO on different threads it will blow up... for TCP this is inadvisable anyway, but for UDP, it can be a valid technique.

An example of this would be `STDIN.read` from multiple threads... not sure how to handle this without adding more layers of indirection.

ioquatix:
By the way, the reason why we cache the wrapper is because it may not be possible to create more than one wrapper per file descriptor with libev - i.e. adding the same file descriptor twice might cause some issues. The wrapper helps track this, but maybe you are right, it's just a limitation of `nio4r` which needs a wrapper allocated per IO.

ioquatix:
I think you are right the only obvious solution right now is to create the wrappers on demand.

PR #106, opened 2021-03-19
Fix wrapper access in the scheduler

bryanp:
## Description

I've noticed this error lately in my recent use of `async-io`:

```
/Users/bryanp/.gem/ruby/3.0.0/gems/async-1.28.9/lib/async/scheduler.rb:74:in `ensure in io_wait': undefined method `reactor=' for nil:NilClass (NoMethodError)
	from /Users/bryanp/.gem/ruby/3.0.0/gems/async-1.28.9/lib/async/scheduler.rb:74:in `io_wait'
	from /Users/bryanp/.rubies/ruby-3.0.0/lib/ruby/3.0.0/socket.rb:304:in `__sendmsg'
	from /Users/bryanp/.rubies/ruby-3.0.0/lib/ruby/3.0.0/socket.rb:304:in `sendmsg'
```

It's been quite tough to reproduce on demand but happens enough I thought it warranted bringing up for discussion.

### Types of Changes

- Bug fix.

### Testing

### DIFF
--- a/lib/async/debug/monitor.rb
+++ b/lib/async/debug/monitor.rb
@@ -20,25 +20,23 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+require 'delegate'
+
 module Async
 	module Debug
-		class Monitor
+		class Monitor < Delegator
 			def initialize(monitor, selector)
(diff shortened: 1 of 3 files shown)


ioquatix:
We have completely removed the need for this in the current implementation. For the current `stable-v1` branch, this may still be applicable.

ioquatix:
Thanks for your contribution.

PR #122, opened 2021-07-24
Add: Add Task#inspect

sevenc-nanashi:
## Description

Task#inspect is so long like:
image

So I made them short.

### Types of Changes

- New feature.

### Testing

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -226,7 +226,7 @@ module Async
 		end
 		
 		def inspect
-			"#<#{self.class} #{@annotation ? "@annotation=\"#{@annotation}\" " : ""}@status=#{@status.inspect}>"
+			"#<#{self.class} @annotation=#{@annotation.inspect} @status=#{@status.inspect}>"
 		end
 		
 		private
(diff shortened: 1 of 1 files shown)


bruno-:
I can confirm inspecting Async object's status has sometimes been a struggle for me too.

ioquatix:
I agree with your assessment, and I already changed it a few weeks ago on main branch. Can you check it?

ioquatix:
https://github.com/socketry/async/blob/7801c07cf46de8cea50ca7a0e3dc0889722de84b/lib/async/node.rb#L220-L240

I don't think this was backporte to `stable-v1` which we could do.

bruno-:
Sorry for the wait. I checked this and can confirm that `main` branch currently contains everything this PR proposes - and more.

`Task#to_s` also has `@status`

https://github.com/socketry/async/blob/731e93bca067a8ee20023bf52e992657120a7500/lib/async/task.rb#L86-L88

So, I think this PR can be closed as this feature is already implemented.

> I don't think this was backported to `stable-v1` which we could do.

I personally can live without this knowing there is an upgrade path.

ioquatix:
I'll consider back porting it. There is a lot I want to do in this space - I want to move fiber/task annotations directly into `Fiber` and `Thread`, but this will take some time.

PR #129, opened 2021-10-13
Use def_delegator instead of alias in Queue

zhulik:
## Description

Replace `alias` with `def_delegator` in `Queue`.

Problem: `Async::LimitedQueue#<<` does not block when the queue is full.

Explanation: both `alias` and `alias_method` despite their misleading names do not create aliases, they create full copies of methods with new names. Therefore if one overrides aliased method in a subclass, the alias name still points to the copy of the original base class method. There are 2 ways of dealing with it:
- defining aliases manually `def new_name(*args, **params) = old_name(*args, **params)`
- using `Forwardable.def_delegator`

Check out a [gist](https://gist.github.com/zhulik/1df6b56d02b2e11f10a28e851da50cd6) and a [blog post](https://theinternate.com/2014/02/14/inheritable-aliases-in-ruby.html)

I'd also recommend replacing aliases everywhere where a class with aliases may potentially be inherited.

Thank you!

### Types of Changes

- Bug fix.

### Testing

- [X] I tested my changes locally.

### DIFF
--- a/lib/async/queue.rb
+++ b/lib/async/queue.rb
@@ -20,12 +20,16 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+require 'forwardable'
+
 require_relative 'notification'
 
 module Async
 	# A queue which allows items to be processed in order.
 	# @public Since `stable-v1`.
 	class Queue < Notification
+		extend Forwardable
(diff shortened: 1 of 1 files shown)


ioquatix:
Ohhh this is a terrible bug lol. Okay, I'll fix it.

ioquatix:
I don't want to introduce forwardable so I've just manually created a proper alias.

PR #144, opened 2022-02-04
add option to raise standard errors

wjwatkinson:
## Description
Give users the option to have their tasks operate like the rest of Ruby and have all errors, including `StandardError`, raised without explicitly having to `wait`, or call `result`.

### Types of Changes

- New feature.

### Testing

### DIFF
--- a/guides/getting-started/README.md
+++ b/guides/getting-started/README.md
@@ -192,6 +192,20 @@ Async do
 end
 ~~~
 
+You can also specify that all unhandled exceptions including `StandardError` be raised immediately using the `:propagate_errors` option. If this option is set in a subtask it will apply to the parent task.
+
+~~~ruby
+require 'async'
+
+Async do |task|
+	task.async(raise_errors: true) do
+		raise "Boom"
+	end
+end
+
+# raises RuntimeError: Boom
+~~~
+
 ## Timeouts
 
 You can wrap asynchronous operations in a timeout. This ensures that malicious services don't cause your code to block indefinitely.
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -62,9 +62,11 @@ module Async
 		# Create a new task.
 		# @parameter reactor [Reactor] the reactor this task will run within.
 		# @parameter parent [Task] the parent task.
-		def initialize(parent = Task.current?, finished: nil, **options, &block)
+		def initialize(parent = Task.current?, raise_errors: false, finished: nil, **options, &block)
 			super(parent, **options)
-			
+
+			@raise_errors = raise_errors
+
 			@status = :initialized
 			@result = nil
 			@finished = finished
@@ -261,7 +263,7 @@ module Async
 				rescue Stop
 					stop!
 				rescue StandardError => error
-					fail!(error, false)
+					fail!(error, @raise_errors)
 				rescue Exception => exception
 					fail!(exception, true)
 				ensure
--- a/spec/async/task_spec.rb
+++ b/spec/async/task_spec.rb
@@ -90,7 +90,29 @@ RSpec.describe Async::Task do
 				end.to raise_exception RuntimeError, /boom/
 			end
 		end
-		
+
+		it "can raise errors" do
+			expect do
+				Async do |task|
+					task.async(raise_errors: true) do |task|
+						raise "boom"
+					end
+				end
+			end.to raise_exception RuntimeError, /boom/
+		end
+
+		it "can raise nested errors" do
+			expect do
+				Async do |task|
+					task.async do |task|
+						task.async(raise_errors: true) do
+							raise "boom"
+						end
+					end
+				end
+			end.to raise_exception RuntimeError, /boom/
+		end
+
 		it "can raise exception after asynchronous operation" do
 			task = nil
 			


ioquatix:
What is the use case?

The problem is, after the task is resumed from the event loop, the exception will bubble up and kill the entire event loop. Is that what you want?

wjwatkinson:
Thanks for the quick reply and, yes, that is what I intended (also apologies for the brief description. I have updated it with more information).

The goal is to create an experience more akin to syncronous ruby, where exceptions are raised and kill the execution unless they are explicitly rescued. [This comment seems to describe the behavior I am looking for, but it is no longer exposed as an option](https://github.com/socketry/async/blob/main/lib/async/task.rb#L231).

An example use case would be if you are aggregating data from multiple APIs and if any of the APIs fail you cannot proceed. Currently you would have to:
```
async = Async do |task|
  req1 = task.async do
     # request 1
  end

  req2 = task.async do
     # request 2
  end

  req1.wait
  req2.wait
end

async.result
```

This feels a bit awkward and verbose. It can also lead to the code taking longer to exit. If, for example req1 takes a long time to complete and req2 errors out immediately the main thread will not error until req1 has finished instead of right after req2 errors.

emiltin:
Elixir (and Erlang) lets you choose how you start jobs (taks). For example, you can choose to link them so that if one crashes, the other one is terminated, or you can keep them isolated. You also have supervisor trees, where a supervisor (parent task) can be used to restart jobs (sub tasks) if they crash.

I don't know whether async should offer more options like that. But I think the current approach of isolating tasks to some degree is good. Concurrent code is, by nature, a bit different. One of the basic ideas of concurrency is that things should run side-by-side. To me that implies some degree of isolation. It make sense to me that a crash in one task place should not cause everything else to fail by default.

ioquatix:
@wjwatkinson I don't think your proposed change works the way you expect. @emiltin gives a good overview of the general concept, but more specifically, I don't think it's reasonable to kill the entire event loop.

That being said, maybe we can achieve want you want in a better way using some helper code for a fan out + exception handling.

ioquatix:
By the way, tasks that raise unhandled errors will log them so if no one calls `wait` it will not be missed entirely.

wjwatkinson:
Thanks for the thoughtful responses. My solution is definitely a bit hacky and killing the entire event loop may be overkill. That said what I struggled with here was how easy and intuitive this package felt until I started testing error handling.

The current design is essentially:
```
rescue StandardError => e
  Console.logger.error(e)
end
```
Which is not something I would do in my code. We also use structured logging (JSON logs), so if I did want to log these errors I would want to do it in a different way.

I think putting the user back in control a bit here and having defaults that align more with synchronous ruby code, if possible , in terms of choosing to rescue instead of choosing to raise, would help. If killing the loop on error is problematic this could perhaps be some sort of option that automatically calls `task.wait` and `async.result`. Users could then rescue and log `StandardError`s if they wanted to.

ioquatix:
> Which is not something I would do in my code. We also use structured logging (JSON logs), so if I did want to log these errors I would want to do it in a different way.

Console supports structured output just redirect stderr to a file and you get newline delimited JSON.

> I think putting the user back in control a bit here and having defaults that align more with synchronous ruby code, if possible , in terms of choosing to rescue instead of choosing to raise, would help. If killing the loop on error is problematic this could perhaps be some sort of option that automatically calls task.wait and async.result.

You can use `Sync` for this behaviour but you won't get fanout.

emiltin:
Is Async::Barrier similar to the kind of fan out operation you want, except that you want any one child failing to abort all children?

https://socketry.github.io/async/source/Async/Barrier/index.html#Async%3A%3ABarrier%23async

ioquatix:
> For my use case, of wanting to make multiple http requests in parallel, that all need to succeed to continue, it would be ideal to have the errors automatically raised in the main thread by default, so they can be handled appropriately.

This is a typical fan out best handled by explicit tasks or a barrier.

You must at some point collect the results of these concurrent requests, no? That's the point where you'd handle exceptions. If you want to have a side channel to cancel the entire operation if one fails, a barrier makes perfect sense.

ioquatix:
With the following example program:

```ruby
#!/usr/bin/env ruby

require_relative '../lib/async'

RAISE_ERRORS = ENV['RAISE_ERRORS'] == 'true'

Async do |task|
	task.async do |task|
		while true
			task.sleep 1
			puts "Hello"
		end
	end
	
	task.async(raise_errors: RAISE_ERRORS) do |task|
		task.sleep 2
		raise "Boom"
	end
end
```

We can check the difference.

image

With the latter, the entire program dies.

ioquatix:
I think a better mechanism is the `Sync do...end` block which is strictly sequential including exception handling. What are the valid use cases for this new feature?

wjwatkinson:
Sorry for the slow reply here. My use case is getting data from multiple sources where all of the data is require to proceed, so if one of the requests fails an exception should be raised. I am going to close this PR out for now as I have not been actively working on this and don't want to clutter the repo.

ioquatix:
For that use case you should definitely just use map-reduce and `Async::Barrier` to handle exceptions.

PR #145, opened 2022-02-05
Modify `yield timer` to `yield timeout` in `Scheduler#timeout_after`

kudojp:
### Why

In the current implementation of `Async::Scheduler#timeout_after`, it executes the given block with the **timer** as the block argument as below.

```rb
    yield timer
```
(from. https://github.com/socketry/async/blob/v2.0.0/lib/async/scheduler.rb#L294)

However, in ruby 3.1, the scheduler which is set to the thread is triggered as below.

```rb
  def timeout(sec, klass = nil, message = nil, &block)   #:yield: +sec+
    # some lines are ommited here.

    if Fiber.respond_to?(:current_scheduler) && (scheduler = Fiber.current_scheduler)&.respond_to?(:timeout_after)
      return scheduler.timeout_after(sec, klass || Error, message, &block)
    end
```
(from. https://github.com/ruby/ruby/blob/ruby_3_1/lib/timeout.rb#L88-L90)

When both `Timeout#timeout` and `Timeout.timeout` is called, **they take the block argument as timeout(second), not as a Timer object**. This is also described in the official doc as below.

> timeout(sec, klass = nil, message = nil) { |sec| ... }
> 
> from. https://docs.ruby-lang.org/en/3.1/Timeout.html

This should be fixed.

### What

`Async::Scheduler#timeout_after` executes the given block with passing timeout as the argument.

```rb
    yield timeout
```

### Types of Changes

- Bug fix.

### Testing

### DIFF
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -291,7 +291,7 @@ module Async
 				end
 			end
 			
-			yield timer
+			yield timeout
 		ensure
 			timer.cancel if timer
 		end


kudojp:
wip


This could be problem.

with_timeout internally calls Fiber.scheduler.timeout_after as below.

```rb
		# @deprecated Replaced by {Scheduler#timeout_after}.
		def with_timeout(timeout, exception = TimeoutError, message = "execution expired", &block)
			Fiber.scheduler.timeout_after(timeout, exception, message, &block)
		end
```
from. https://github.com/kudojp/async/blob/main/lib/async/task.rb#L95-L98


```rb
	describe '#with_timeout' do
		it "can extend timeout" do
			reactor.async do |task|
				task.with_timeout(0.2) do |timer|
					task.sleep(0.1)
					
					expect(timer.fires_in).to be_within(10 * Q).percent_of(0.1)
					
					timer.reset
					
					expect(timer.fires_in).to be_within(10 * Q).percent_of(0.2)
				end
			end
			
			reactor.run
		end
```

https://github.com/kudojp/async/blob/main/spec/async/task_spec.rb#L368

kudojp:
Look for a place where timeout_after is used.

## rg timeout_after

```rb
$ rg timeout_after
lib/async/task.rb
95:             # @deprecated Replaced by {Scheduler#timeout_after}.
97:                     Fiber.scheduler.timeout_after(timeout, exception, message, &block)

lib/async/reactor.rb
37:             alias with_timeout timeout_after

lib/async/scheduler.rb
285:            def timeout_after(timeout, exception = TimeoutError, message = "execution expired", &block)
```

### 1. lib/async/task.rb L95-L98

```rb
		# @deprecated Replaced by {Scheduler#timeout_after}.
		def with_timeout(timeout, exception = TimeoutError, message = "execution expired", &block)
			Fiber.scheduler.timeout_after(timeout, exception, message, &block)
		end
```

### 2. lib/async/reactor.rb L37



### 3. lib/async/scheduler.rb L285

kudojp:
I wonder if I could the feature to reset/rescheduler timer of the task which is scheduled by `Task#with_timeout`.

Since Task#with_timeout is now deprecated,  it is a possible idea to drop this feature.

kudojp:
Mmm,,,

Now I have found that with this PR, resetting/rescheduling timer of the task which is scheduled by Scheduler#timeout_after gets also impossible. I may need a good interface to enable it.

kudojp:
I would like you to ask you as the owner of this repo a question as the own.



In this PR, I updated Async::Scheduler#timeout_after so that Async::Scheduler implements the FiberScheduler as specified. However, with this change,  the interface to reset/reschedule the tasks which is scheduled with this method would be lost as below.

<details>
<summary>details</summary>

In the current interface,,,

```rb
Fiber.schedule do
	timeout.timeout(10) do |t| # t is Timers::Timer object
		sleep(5)
		t.reset()
	end
end
```

</details>


After the update of this PR, it is impossible.

```rb
Fiber.schedule do
	timeout.timeout(10) do |t| # t is timeout(sec)
		sleep(20)
	end
end
```

</details>

There are some possible solutions.

### Solution1. Give up to implement FiberScheduler interface

For that, we just have to close this PR.

Pros.
- No job to be done.

Cons.
- The scheduler does not implement an FiberScheduler interface, which is not expected.


### Solution2. Implement a FiberScheduler interface and accept to lose capability of reseting/rescheduling tasks

For this, we just have to merge this PR.

Pros.
- The scheduler does not implement an FiberScheduler interface as expected.

Cons.
- Some users may reset/reschedule tasks now and they cannot do it with the next version of this gem.


### Solution3. Implement FiberScheduler interface and add another interface to reseting/rescheduling tasks

For this, we have to update this PR even more.

Pros.
- The scheduler does not implement an FiberScheduler interface as expected.

Cons.
- It takes time since we have to think from the design.
- If the interface for that is defined FiberScheduler in the future, we have to implement again.

I personally think this is the best way. I think if you think that the capability to reseting/rescheduling tasks, I think updating the interface of FiberScheduler should comes first. If Ruby maintainers think that reseting/rescheduling is useful, I think this gem should also implement it.

ioquatix:
Thanks for all the details and notes. I will implement option 3 or some variant of it as the interface should be compatible between Async 1 and 2. I have a locally working branch I'll push it later this week.

ioquatix:
Thanks I have merged a fix for this.

kudojp:
Thanx!

ioquatix:
I will try to release this soon after I test the changes.

PR #149, opened 2022-02-18
Ensure timeout_after yields duration.

ioquatix:
## Description

As discussed in https://github.com/socketry/async/pull/145

We need to retain backwards compatibility with Async 1 interface `with_timeout` which is an acceptable middle ground.

### Types of Changes

- Bug fix.
- Breaking change.

### Testing

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -92,9 +92,9 @@ module Async
 			super
 		end
 		
-		# @deprecated Replaced by {Scheduler#timeout_after}.
-		def with_timeout(timeout, exception = TimeoutError, message = "execution expired", &block)
-			Fiber.scheduler.timeout_after(timeout, exception, message, &block)
+		# Execute the given block of code, raising the specified exception if it exceeds the given duration during a non-blocking operation.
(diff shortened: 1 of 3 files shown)


lib/async/task.rb
```diff
@@ -92,9 +92,9 @@ def sleep(duration = nil)
 			super
 		end
 		
-		# @deprecated Replaced by {Scheduler#timeout_after}.
-		def with_timeout(timeout, exception = TimeoutError, message = "execution expired", &block)
-			Fiber.scheduler.timeout_after(timeout, exception, message, &block)
+		# Execute the given block of code, raising the specified exception if it exceeds the given duration during a non-blocking operation.
+		def with_timeout(duration, exception = TimeoutError, message = "execution expired", &block)
+			Fiber.scheduler.with_timeout(timeout, exception, message, &block)
```
olleolleolle, line 97: Perhaps duration/timeout is a typo here. Failed the tests, I guess?

ioquatix: Yes! Thanks :)

PR #163, opened 2022-05-04
Fix handling of barriers when waiting on a task fails with an unrelated exception.

ioquatix:
When using a barrier with a timeout, there was a bug where the task would be removed from the task list even though it was still running, causing `Barrier#stop` to later miss stopping the task and leak it.

### Types of Changes

- Bug fix.

### Testing


ioquatix:
Merged by hand.

PR #170, opened 2022-06-15
Log as warning rather than error, and with explicit message.

ioquatix:
At best these errors are harmless and at worst they represent a logic error in the program. Since we expect this to mostly be a false positive, change it to a warning.

### Types of Changes

- New feature.


lib/async/task.rb
```diff
@@ -224,18 +224,19 @@ def complete?
...
-			elsif @finished.nil?
-				# If no one has called wait, we log this as an error:
-				Console.logger.error(self) {$!}
-			else
-				Console.logger.debug(self) {$!}
+			if exception
+				if propagate
+					raise exception
+				elsif @finished.nil?
+					# If no one has called wait, we log this as an error:
```
olleolleolle, line 235:
```suggestion
					# If no one has called wait, we log this as a warning:
```

Issue #62, opened 2020-04-03
Add support to Async::Barrier to wait for the first response


tadman:
There are occasions where any one of several Task activities might produce a result and only the first result matters.

Ideally there's a method like `#wait_first` which can return the result of the first task to finish.

In some cases it may be necessary to select the first non-`nil` result, but that could be handled with an `Async::Condition` instead.

zhulik:
Just in case if someone is looking for an implementation:

```ruby
# frozen_string_literal: true

def wait_first(*tasks, task: Async::Task.current)
  c = Async::Notification.new

  await = lambda do |task|
    task.async do
      task.wait
      c.signal(task)
    rescue StandardError
      c.signal(task)
    end
  end

  tasks.each(&await)

  first = c.wait

  [first, tasks - [first]]
end
```

Usage:
```ruby
done, pending = wait_first(task1, task2, task3)
pending.each(&:stop) # cancel others
done.wait # return or re-raise first result
```

The API is influenced by python's `asyncio.wait(tasks,return_when=asyncio.FIRST_COMPLETED)`

Personally I'd be happy to have a similar method defined on `Async`. Something like
`Async.wait(*tasks, mode: :all)` - awaits all tasks, a shortcut for `Barrier`, returns
`Async.wait(*tasks, mode: :first)` - awaits first, either my implementation or a shortcut for some primitive with similar logic

I'd like to volunteer myself for this task. WDYT?

ioquatix:
Sure sounds great, I don't have any strong opinions on the interface except that if it's fundamentally different from `wait` we should make a separate method.

zhulik:
What about `Async.wait_for`?

ioquatix:
Waiting for all tasks vs waiting for the first task are quite different things IMHO and the use case is different. Waiting for all tasks can be achieved by `tasks.each(&:wait)`. Waiting for the first N tasks to complete is more complex since you need to track all the tasks.

Needing to spawn N tasks to wait for the first task to finish out of N tasks seems like an anti-pattern.

We might need to consider how to do this more efficiently.

zhulik:
A Barrier-like primitive?

```ruby
w = Async::Waiter.new # naming is currently out of scope
w.async{}
w.async{}
w.async{}
done, pending = w.wait_first(2) # or maybe w.wait(first: 2). Wait without arguments will wait for all just like Barrier, but with different return value
```

ioquatix:
The most efficient implementation is something like this:

```ruby
[...]

PR #174, opened 2022-07-30
Add Async::Waiter

zhulik:
**WIP**

As discussed [here](https://github.com/socketry/async/issues/62) I introduce a new class called `Waiter` which can be a drop-in replacement  `Barrier`. The only difference is that it can await for first N tasks.

Once and if the approach is approved I'll add more specific tests

Questions:
- Better name?
- Replace Barrier?
- How to get rid of warnings in logs when a task raises an exception? See example:
```ruby
Async do |task|
  waiter = Async::Waiter.new
  waiter.async do
    task.sleep(1)
    raise StandardError, "Error!"
  end
  waiter.async do
    task.sleep(2)
    2
  end
  waiter.async do
    task.sleep(3)
    3
  end
  waiter.async do
    task.sleep(3)
    raise StandardError, "Error!"
  end

  done, pending = waiter.wait_for(2)

  done.each do |task|
    Console.logger.info(self, "Done result: #{task.wait}")
  rescue StandardError => e
    Console.logger.info(self, e)
  end

  pending.each do |task|
    Console.logger.info(self, "Pending result: #{task.wait}")
  rescue StandardError => e
    Console.logger.info(self, e)
  end
  Console.logger.info(self, "Done")
end
```

Even though the tasks with raises have eventually been awaited, I still see warnings in the logs:
```
  0.0s     warn: Async::Task [oid=0x8c] [ec=0xa0] [pid=453505] [2022-07-30 23:31:16 +0200]
               | Task may have ended with unhandled exception.
               |   StandardError: Error!
               |   → ./waiter.rb:56 in `block (2 levels) in <main>'
               |     ./waiter.rb:20 in `block in async'
               |     /home/zhulik/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/async-2.0.3/lib/async/task.rb:255 in `block in schedule'
  1.0s     info: Object [oid=0xb4] [ec=0xc8] [pid=453505] [2022-07-30 23:31:17 +0200]
               |   StandardError: Error!
               |   → ./waiter.rb:56 in `block (2 levels) in <main>'
               |     ./waiter.rb:20 in `block in async'
               |     /home/zhulik/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/async-2.0.3/lib/async/task.rb:255 in `block in schedule'
  1.0s     info: Object [oid=0xb4] [ec=0xc8] [pid=453505] [2022-07-30 23:31:17 +0200]
               | Done result: 2
  2.0s     info: Object [oid=0xb4] [ec=0xc8] [pid=453505] [2022-07-30 23:31:18 +0200]
               | Pending result: 3
  2.0s     info: Object [oid=0xb4] [ec=0xc8] [pid=453505] [2022-07-30 23:31:18 +0200]
               |   StandardError: Error!
               |   → ./waiter.rb:68 in `block (2 levels) in <main>'
               |     ./waiter.rb:20 in `block in async'
               |     /home/zhulik/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/async-2.0.3/lib/async/task.rb:255 in `block in schedule'
  2.0s     info: Object [oid=0xb4] [ec=0xc8] [pid=453505] [2022-07-30 23:31:18 +0200]
               | Done
```

**Note**: I'm leaving on vacation on August 2, the day after tomorrow and won't be able to code till August 9.

### Types of Changes

- New feature.

### Testing

### DIFF
--- a/lib/async/waiter.rb
+++ b/lib/async/waiter.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require_relative 'barrier'
+
+module Async
+	class Waiter < Barrier
+		def initialize(parent: nil)
+			super
+			@finished = Async::Condition.new
+			@all_tasks = []
+			@done = []
+		end
+
+		def async(*arguments, parent: (@parent or Task.current), **options)
+			t = parent.async(*arguments, **options) do |task|
+				yield(task)
+			ensure
+				@done << task
+				@finished.signal
+			end
+
+			@tasks << t
+			@all_tasks << t
+			t
+		end
+
+		def wait_for(n = size)
+			raise ArgumentError, "'n' cannot be greater than size. Given: #{n}, size: #{size}" if n > size
+
+			while @done.size < n
+				@finished.wait
+			end
+
+			done = @done.first(n)
+			[done, @all_tasks - done]
+		ensure
+			@tasks.filter!{ |t| !@done.include?(t) }
+		end
+
+		def wait(n = size)
+			wait_for(n).first.map(&:wait)
+		end
+	end
+end
--- a/spec/async/barrier_examples.rb
+++ b/spec/async/barrier_examples.rb
@@ -1,0 +1,177 @@
+# frozen_string_literal: true
+
+require 'async/clock'
+require 'async/rspec'
+require 'async/semaphore'
+
+require_relative 'chainable_async_examples'
+
+RSpec.shared_examples 'barrier' do
+	describe '#async' do
+		let(:repeats) {40}
+		let(:delay) {0.1}
+
+		it 'should wait for all jobs to complete' do
+			finished = 0
+
+			repeats.times.map do |i|
+				subject.async do |task|
+					task.sleep(delay)
+					finished += 1
+
+					# This task is a child task but not part of the barrier.
+					task.async do
+						task.sleep(delay*3)
+					end
+				end
+			end
+
+			expect(subject).to_not be_empty
+			expect(finished).to be < repeats
+
+			duration = Async::Clock.measure{subject.wait}
+
+			expect(duration).to be < (delay * 2 * Q)
+			expect(finished).to be == repeats
+			expect(subject).to be_empty
+		end
+	end
+
+	describe '#wait' do
+		it 'should wait for tasks even after exceptions' do
+			task1 = subject.async do
+				raise "Boom"
+			end
+
+			task2 = subject.async do
+			end
+
+			expect(task1).to be_failed
+			expect(task2).to be_finished
+
+			expect{subject.wait}.to raise_exception(/Boom/)
+
+			subject.wait until subject.empty?
+
+			expect(subject).to be_empty
+		end
+
+		it 'waits for tasks in order' do
+			order = []
+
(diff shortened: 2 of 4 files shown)


spec/async/barrier_spec.rb
```diff
@@ -1,17 +1,17 @@
 # frozen_string_literal: true
 
 # Copyright, 2019, by Samuel G. D. Williams. <http://www.codeotaku.com>
-# 
+#
```
zhulik, line 4: What would you say if I start slowly introducing Rubocop?
I've already tried, got about 7.5k(!) issues, .rubocop_todo.yml is about 700 lines long.

I could start with simple cases likes this, missing newlines at the end of files or enforcing tabs for indentation(no need to fix, code is already properly formatted afair)

spec/async/barrier_examples.rb
```diff
@@ -0,0 +1,177 @@
+# frozen_string_literal: true
+
+require 'async/clock'
+require 'async/rspec'
+require 'async/semaphore'
```
zhulik, line 5: What do you think about zeitwerk?

lib/async/waiter.rb
```diff
@@ -0,0 +1,44 @@
...
+			[done, @all_tasks - done]
+		ensure
+			@tasks.filter!{ |t| !@done.include?(t) }
+		end
+
+		def wait(n = size)
+			wait_for(n).first.map(&:wait)
+		end
+	end
+end
```
zhulik, line 44: See commend about Rubocop

spec/async/barrier_examples.rb
ioquatix: Autoload is broken on Ruby with the fiber scheduler so it's a no go for now, but maybe later?

spec/async/barrier_spec.rb
ioquatix: haha, I misread that as 75k issues lol.

When I tried RuboCop a while ago, IIRC it did not support tabs for indentation.

However, I believe that changed recently.

I'd be okay with introducing RuboCop but it would probably need to be the opposite way for me personally, it would need to disable every rule by default and we selectively enable ones that make sense. It would probably just be a small subset and if I did go in that direction I'd want to replicate it to other projects too.

ioquatix:
This looks good to me - I like using a sub-class since you need to opt into it - there is extra overhead. I will require people know ahead of time they want to wait on a subset of items. Is there any other design we should consider?

I wonder if it's worth looking at how other implementations work and how they handle this particular overhead.

bruno-:
From #62:

> The idea of a barrier is that it's stateful, you can call wait(2) several times until it's exhausted.

Does the current implementation enable this feature? Since `@done` never changes I think `#wait(2)` will always return the same result, no?

zhulik:
@bruno- Waiter passes all tests written for Barrier. At least for the case when all task are being awaited. For the case when N tasks are being awainted I decided that it's better when all coroutines waiting for a waiter to receive same N first results

ioquatix:
Closing in favour of #189 as discussed.

Issue #176, opened 2022-09-01
Memory leak when using Task#with_timeout


jpaulgs:
```
#!/usr/bin/env ruby

require 'async'
require 'async/queue'

@input_channel = Async::LimitedQueue.new(100)

def consumer
  Async do |task|
    loop do
      task.with_timeout(0.001) do
        @input_channel.dequeue
      rescue Async::TimeoutError
        # no-op
      end
    end
  end
end

task = consumer

task.wait
```

Hey, I'm running ruby 3.1.2 and version 2.0.3 of the Async rubygem.  The code above is similar to what we are using. It exhibits a memory leak.  The following photo was from running this for about 15 minutes, it certainly never releases the memory.

![image](https://user-images.githubusercontent.com/30516993/188029733-a3272fa9-34e6-436a-9b58-002c28cfaa6f.png)

I'll keep digging to see what I can find out.

jpaulgs:
Actually the LimitedQueue is irrelevant here, the following code also shows the issue

```
#!/usr/bin/env ruby

require 'async'

def consumer
  Async do |task|
    loop do
      task.with_timeout(0.001) do
        # no-op
      rescue Async::TimeoutError
        # no-op
      end
    end
  end
end

task = consumer

task.wait
```

![image](https://user-images.githubusercontent.com/30516993/188030393-806622bb-5c80-4dae-97d9-aeb731c14d5b.png)

ioquatix:
Hmm, thanks for reporting this, I'll take a look.

ioquatix:
``` ruby
#!/usr/bin/env ruby

require 'async'
require 'memory'

def consumer
	Async do |task|
		report = Memory.report do
			1000000.times do
				task.with_timeout(0.001) do
					# no-op
				rescue Async::TimeoutError
					# no-op
				end
			end
		end
		report.print
		binding.irb
	end
end

task = consumer

task.wait
```

# Memory Profile

- Total Allocated: (28.00 MiB in 4000000 allocations)
- Total Retained: (7.00 MiB in 1000000 allocations)

## By Gem (288.00 MiB in 4000000 allocations)

- (288.00 MiB in 4000000 allocations)	timers-4.3.3

## By File (288.00 MiB in 4000000 allocations)

- (160.00 MiB in 2000000 allocations)	/home/samuel/.gem/ruby/3.1.2/gems/timers-4.3.3/lib/timers/timer.rb
[...]

PR #186, opened 2022-10-27
Use Set instead of Array for Condition @waiting queue to avoid memory leak

oeoeaio:
Related to #176 

We found that calling `dequeue` on instances of `Queue` (and `LimitedQueue`) inside a `Task.with_timeout` block inside a loop will result in fibers accumulating in the `@waiting` list for the Queue in the absence of any other fiber signalling the `Queue`, leading to a memory leak.

Our short term fix for this was just to manually `signal` the `Queue` whenever the timeout was reached, but this only works where a single fibre is waiting on the queue with a timeout in a loop.

Switching to using a `Set` instead of an `Array` to hold the list of waiting fibres resolves the core issue. I don't think this will break anything as I can't really see a use case for having the same fibre in the `@waiting` list multiple times.

Keen for input on what tests might be appropriate for this, if any. Is suppose the main thing is that this change does not break the existing test suite.

## Types of Changes
- Bug fix.

## Contribution

## Profiling code
```ruby
#!/usr/bin/env ruby

require 'async'
require 'async/queue'
require 'memory_profiler'

def produce(input_channel)
  Async do
    input_channel.enqueue('some infrequent message')
  end
end

def consume(input_channel)
  Async do |task|
    buffer = []
    report = MemoryProfiler.report do
      10000.times do
        # This timeout is artificially tight to demonstrate the memory leak
        task.with_timeout(0.001) do
          buffer << input_channel.dequeue
        rescue Async::TimeoutError
          # do something with the buffer
        end
        # do something with the buffer
      end
    end
    report.pretty_print
  end
end

Async do
  input_channel = Async::Queue.new
  produce(input_channel)
  consume(input_channel)
end
```

## Profiling before this change
retained memory by location
-----------------------------------
    399960  /usr/local/bundle/gems/async-2.2.1/lib/async/condition.rb:37
       584  /usr/local/bundle/gems/async-2.2.1/lib/async/scheduler.rb:73
       160  /usr/local/bundle/gems/timers-4.3.5/lib/timers/timer.rb:25
        88  /usr/local/bundle/gems/timers-4.3.5/lib/timers/group.rb:46
        40  /usr/local/bundle/gems/async-2.2.1/lib/async/scheduler.rb:273
        40  /usr/local/bundle/gems/timers-4.3.5/lib/timers/events.rb:62

retained memory by class
-----------------------------------
    399960  Async::Condition::Queue
       584  Thread::Backtrace
       160  Proc
        88  Timers::Timer
        40  Async::TimeoutError
        40  Timers::Events::Handle

## Profiling after this change
retained memory by location
-----------------------------------
       584  /usr/local/bundle/gems/async-2.2.1/lib/async/scheduler.rb:73
       384  /usr/local/lib/ruby/3.1.0/set.rb:522
       192  /usr/local/lib/ruby/3.1.0/set.rb:540
       160  /usr/local/bundle/gems/timers-4.3.5/lib/timers/timer.rb:25
        88  /usr/local/bundle/gems/timers-4.3.5/lib/timers/group.rb:46
        40  /usr/local/bundle/gems/async-2.2.1/lib/async/scheduler.rb:273
        40  /usr/local/bundle/gems/timers-4.3.5/lib/timers/events.rb:62

retained memory by class
-----------------------------------
       584  Thread::Backtrace
       576  Hash
       160  Proc
        88  Timers::Timer
        40  Async::TimeoutError
        40  Timers::Events::Handle

### DIFF
--- a/lib/async/condition.rb
+++ b/lib/async/condition.rb
@@ -12,20 +12,16 @@ module Async
 	# @public Since `stable-v1`.
 	class Condition
 		def initialize
-			@waiting = []
+      @waiting = Set.new
 		end
 		
 		Queue = Struct.new(:fiber) do
 			def transfer(*arguments)
-				fiber&.transfer(*arguments)
+				fiber.transfer(*arguments)
 			end
 			
 			def alive?
-				fiber&.alive?
-			end
-			
-			def nullify
-				self.fiber = nil
+				fiber.alive?
 			end
 		end
 		
@@ -35,11 +31,11 @@ module Async
 		# @returns [Object]
 		def wait
 			queue = Queue.new(Fiber.current)
-			@waiting << queue
-			
+      @waiting << queue
+
 			Fiber.scheduler.transfer
 		ensure
-			queue.nullify
+      @waiting.delete(queue)
 		end
 		
 		# Is any fiber waiting on this notification?
@@ -51,9 +47,9 @@ module Async
 		# Signal to a given task that it should resume operations.
 		# @parameter value [Object | Nil] The value to return to the waiting fibers.
 		def signal(value = nil)
-			waiting = @waiting
-			@waiting = []
-			
+      waiting = @waiting.dup
+      @waiting.clear
+
 			waiting.each do |fiber|
 				Fiber.scheduler.resume(fiber, value) if fiber.alive?
 			end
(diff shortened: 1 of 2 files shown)


ioquatix:
The ideal solution for this is to use a linked list. Short of a native implementation, we could fall back to using a native Ruby mutex/queue. It might even be faster, but the semantics aren’t quite the same. A Ruby mutex uses a linked list internally and my goal was to preserve FIFO order. I like your PR but I think we might need to be a little more sensitive to order and I’m not sure about performance either.

Do you mind opening another PR using Thread::Queue as the queue? Let’s see if it works.

PR #189, opened 2022-10-30
Add a concurrency primitive for waiting for a specific number of tasks to complete.

ioquatix:
See <https://github.com/socketry/async/pull/174> for a slightly different implementation.

@bruno- @zhulik what do you think of this? You can compose it with barrier if you need a final wait-all, which simplifies the implementation and also makes it composable with semaphore, etc.

### DIFF
--- a/lib/async/waiter.rb
+++ b/lib/async/waiter.rb
@@ -1,0 +1,33 @@
+
+module Async
+	# A composable synchronization primitive, which allows one task to wait for a number of other tasks to complete. It can be used in conjunction with {Semaphore} and/or {Barrier}.
+	class Waiter
+		def initialize(parent: nil, finished: Async::Condition.new)
+			@finished = finished
+			@done = []
+			
+			@parent = parent
+		end
+		
+		def async(parent: (@parent or Task.current), &block)
+			parent.async do |task|
+				yield(task)
+			ensure
+				@done << task
+				@finished.signal
+			end
+		end
+		
+		def wait_for(count)
+			while @done.size < count
+				@finished.wait
+			end
+			
+			return @done.shift(count)
+		end
+		
+		def wait(count, exception: false)
+			wait_for(count).map(&:result)
+		end
+	end
+end
--- a/lib/async/waiter.rb
+++ b/lib/async/waiter.rb
@@ -26,7 +26,7 @@ module Async
 			return @done.shift(count)
 		end
 		
-		def wait(count, exception: false)
+		def wait(count)
 			wait_for(count).map(&:result)
 		end
 	end
--- a/lib/async/limited_barrier.rb
+++ b/lib/async/limited_barrier.rb
@@ -1,7 +1,11 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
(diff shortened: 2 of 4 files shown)


ioquatix:
I'm not sure if the name of this is class is correct. `Async::Waiter` just seems too generic.

If you only care about waiting for some subset of tasks, this + a semaphore is probably more useful than a barrier. I think they serve different purposes.

lib/async/waiter.rb
```diff
@@ -0,0 +1,33 @@
...
+		
+		def wait_for(count)
+			while @done.size < count
+				@finished.wait
+			end
+			
+			return @done.shift(count)
+		end
+		
+		def wait(count, exception: false)
```
bruno-, line 29: Is `exception: false` needed?

bruno-:
I like this implementation better 👍

> Async::Waiter just seems too generic

Yes, and also the first association for waiter is that is should **serve** something

waiter

Maybe `Async::Moat`?

lib/async/waiter.rb
ioquatix: No, it is a hangover from a previous permutation.

ioquatix:
- `Async::FinishLine`
- `Async::Winners`
- `Async::Deadline`
- `Async::Goal`
- `Async::Race` (used in JS?)
- `Async::Battle`

ioquatix:
For deadline scheduling, it could also be useful:

```ruby
barrier = Async::Barrier.new
waiter = Async::Waiter.new(parent: barrier)
urls.each{|url| waiter.async{... url ...}}
waiter.each(timeout: 5) do
  # Process as many completions as possible within 5 seconds
end
barrier.stop # kill all remaining tasks
```

bruno-:
`Async::LimitedBarrier` sounds the best IMO.

lib/async/waiter.rb
```diff
@@ -0,0 +1,33 @@
...
+	# A composable synchronization primitive, which allows one task to wait for a number of other tasks to complete. It can be used in conjunction with {Semaphore} and/or {Barrier}.
+	class Waiter
+		def initialize(parent: nil, finished: Async::Condition.new)
+			@finished = finished
+			@done = []
+			
+			@parent = parent
+		end
+		
+		def async(parent: (@parent or Task.current), &block)
```
zhulik, line 12: Why do you prefer `or` over `||` here and in other similar places?

zhulik:
I like this implementation and `Async::LimitedBarrier` seems to be a good name IMO

lib/async/waiter.rb
ioquatix: `|` is bitwise or.
`||` is boolean or.
`or` is logical/semantic/object or.

That's my mental model. I try to be consistent but I'm not always. I also prefer `or` when the lhs and rhs are disjoint and you are only concerned with the semantics "execute this or execute that" vs "result = boolean this || boolean that". I think the goal for me is to communicate that this isn't a boolean result we care about (true or false). The fact that `||` doesn't just convert everything to true/false is odd to me.

PR #192, opened 2022-10-31
Use a linked list for the barrier implementation.

ioquatix:
This simplifies the implementation and unifies the semantics with other concurrency primitives that also use a linked list.

It introduces a minor breaking change, in that `Barrier#wait` will no longer raise exceptions. Because previously the code would have been to wrap this in exception handling and retry, this should be objectively safer with no downside for existing correct code, and fix hidden issues in code which does not have retries.

## Types of Changes

- Bug fix.
- New feature.
- Breaking change.

## Contribution

### DIFF
--- a/lib/async/barrier.rb
+++ b/lib/async/barrier.rb
@@ -3,6 +3,7 @@
 # Released under the MIT License.
 # Copyright, 2019-2022, by Samuel Williams.
 
+require_relative 'list'
 require_relative 'task'
 
 module Async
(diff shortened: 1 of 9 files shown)


trevorturk:
I noticed this commit and tested with my app, and it's a breaking change for me. I was relying on `Barrier#wait` to raise exceptions. What's the recommended alternative if I have a barrier and I want exceptions in the barrier's tasks to propagate so that I can rescue and handle them? (Apologies if I was doing this incorrectly before, but I thought this was the way!)

ioquatix:
Yes, if this breaks code we may need to change it back or add an `exception: true/false` argument or a separate method. Thanks for your report.

Can you tell me how you were waiting on existing tasks?

I see two issues

- `Barrier#tasks` is now longer a direct array of tasks - we can fix this.
- `Barrier#wait` no longer raises exceptions.

Did you call `barrier.wait` in a loop?

trevorturk:
(If anyone else is reading this, I sent some code samples to Samuel privately to answer his usage questions.)

It looks like #197 restores the preexisting behavior, and #195 introduces `Async::Group#wait` which also seems interesting for my use cases. I'll keep an eye on these changes and continue to test as things move forward. Thank you!

PR #194, opened 2022-11-02
Better support for `Fiber.set_scheduler`.

ioquatix:
Add support for `scheduler_close` hook which is canonical for `Fiber.set_scheduler` usage, without it, the scheduled fibers won't run unless the scheduler is explicitly invoked.

## Types of Changes

- Bug fix.
- New feature.

## Contribution

### DIFF
--- a/lib/async/reactor.rb
+++ b/lib/async/reactor.rb
@@ -21,6 +21,10 @@ module Async
 			Fiber.set_scheduler(self)
 		end
 		
+		def scheduler_close
+			self.close
+		end
+		
 		public :sleep
 	end
 end
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -34,6 +34,12 @@ module Async
 			@timers = ::Timers::Group.new
 		end
 		
+		def scheduler_close
+			self.run
+		ensure
+			self.close
+		end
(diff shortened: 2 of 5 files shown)


bruno-:
Hi,

I always thought this is handled via `#close` method.
Were there any changes to the `FiberScheduler` interface for 3.2? I checked the messages of your commits to Ruby for the last couple months and I couldn't find anything...

ioquatix:
When I first implemented the fiber scheduler, I called close when going out of scope (end of thread, process, etc).

I found that this was pretty annoying because it's different from when the user calls `scheduler.close`. It was basically a mistake.

So the implementation now looks like this: <https://github.com/ruby/ruby/blob/59a6caf83acf32915abf11ae81ca3167d2577b21/scheduler.c#L210-L240>.

Basically, when going out of scope, it will invoke `scheduler_close` which should run the fiber scheduler and then close it. You can also expose a `scheduler.close` to the user.

In `Async::Reactor#close`, it should NOT run the reactor but instead directly close it. So, `Async::Reactor` as a scheduler won't run the reactor when going out of scope, which is not the expected behaviour.

bruno-:
Thank you for the elaborate explanation

PR #205, opened 2022-12-05
Add guides about barriers with a fan out and termination

trevorturk:
Add a bit more to the guides about how to use Async::Barrier with a fan out of Async::Tasks with proper termination. 

## Types of Changes

- Guides

## Contribution


ioquatix:
This looks good to me. I wonder if we should break this out into a separate document about expected patterns, e.g. fan out, map reduce, first N completed, etc.

PR #236, opened 2023-06-02
Follow up from fix trasient typo

ColinDKelley:
Follow-up to "Fix trasient typo" [PR](https://github.com/socketry/async/pull/235).

## Types of Changes

- DocumenttIon

## Contribution


guides/asynchronous-tasks/readme.md
```diff
@@ -359,44 +382,48 @@ The purpose of transient tasks is when a task is an implementation detail of an
 - A background worker or batch processing job which is independent of any specific operation, and is lazily created.
 - A cache system which needs periodic expiration / revalidation of data/values.
 
-Bearing in mind, in all of the above cases, you may need to validate that the background task hasn't been stopped, e.g.
+Bear in mind, in all of the above cases, you may need to validate that the background task hasn't been stopped, e.g. ???
```
ColinDKelley, line 385: @ioquatix Can you suggest how to finish this sentence where the ??? is? I wasn't sure.

ioquatix: I think we can delete the sentence, IIRC, what I was trying to say, was that such tasks could be stopped due to going out of scope, rather than explicitly being `Task#stop`ed. In other words, `@refresh ||= Async(transient: true) {...}` is required to restart tasks if they go out of scope and have been stopped - "you may need to validate the background task hasn't been stopped".

A more specific example: let's say you had an in-process background job processing system. When you schedule a job, you should ensure it's running when adding the job. Otherwise, it may stop as a matter of program flow, and equally, it must be started as a matter of program flow.

This is captured in the below example.

ColinDKelley: Sentence deleted.

guides/asynchronous-tasks/readme.md
```diff
@@ -359,44 +382,46 @@ The purpose of transient tasks is when a task is an implementation detail of an
...
 	def refresh!
 		@refresh ||= Async(transient: true) do
 			loop do
-				@current_time = Time.now.to_s
+				@current_time_string = Time.now.to_s
 				sleep(1)
 			end
 		ensure
-			# When the reactor terminates all tasks, this will be invoked:
+			# When the reactor terminates all tasks, this will be invoked.
```
ColinDKelley, line 415: @ioquatix So this `ensure` block will be run because the reactor terminating the task will cause an exception to be raised from `sleep(1)`?

BTW do we prefer `sleep(1)` or `sleep 1` in this doc? I see both used here.

ioquatix: > So this ensure block will be run because the reactor terminating the task will cause an exception to be raised from sleep(1)?

Yes that's correct.

> do we prefer sleep(1) or sleep 1

I probably prefer `sleep(1)`.

Also, when comments refer to the next line, I prefer to end with a `:`.

guides/asynchronous-tasks/readme.md
```diff
@@ -52,32 +52,32 @@ At any point in your program, you can start the reactor and a root task using th
...
 	3.times do |i|
-		sleep 1
+		sleep 1 + i/10.0
 		puts "Hello World #{i}"
 	end
 end
 ```
 
-This program prints "Hello World" 3 times. Before printing, it sleeps for 1 second. The total execution time is 3 seconds because the program executes sequentially.
+This program prints "Hello World" 3 times. Before printing, it sleeps for 1, then 1.1, then 1.2 seconds. The total execution time is 3.3 seconds because the program executes sequentially.
```
ioquatix, line 61: Do you mind explaining the motivation to change this to `+ i/10.0`?

guides/asynchronous-tasks/readme.md
```diff
@@ -183,22 +183,24 @@ barrier.wait
 
 ## Stopping a Task
 
-When a task completes execution, it will enter the `complete` state (or the `failed` state if it raises an unhandled exception).
```
ioquatix, line 186: All exceptions will cause the task to enter the failed state.

https://github.com/socketry/async/blob/main/lib/async/task.rb#L314-L317

guides/asynchronous-tasks/readme.md
```diff
@@ -359,44 +382,46 @@ The purpose of transient tasks is when a task is an implementation detail of an
...
 end
 
 Async do
-	# If you are handling 1000s of requests per second, it can be an advantage to cache the current time string.
-	p TimeCache.instance.current_time_string
+	p TimeStringCache.instance.current_time_string
 end
 ```
 
-Upon existing the top level async block, the {ruby @refresh} task will be set to `nil`. Bear in mind, you should not share these resources across threads; doing so would need some form of mutual exclusion.
```
ioquatix, line 402: I'm not sure if we should remove this statement, it should be clear that `@refresh` is bound to the current reactor per thread.

guides/asynchronous-tasks/readme.md
ColinDKelley: Ok, I've updated all uses of `sleep` to `sleep(...)` and terminated the comment (on the next line) with `:`.

guides/asynchronous-tasks/readme.md
ColinDKelley: Yes, I was trying to make the example more clear about when the stop happens. If all of the timers are firing at the 1 second mark, the timing of the stop is harder to follow.

But the tenths of a second are a bit cluttered for sure. I'd be happy to make those whole seconds or use some other approach for each thread to have a different timing value. Or I can put it back as it was if you'd prefer.

guides/asynchronous-tasks/readme.md
ColinDKelley: I thought you were saying that non-`StandardError`-derived exceptions would shut down the reactor and cause it to re-`raise`?

guides/asynchronous-tasks/readme.md
ioquatix: I see, I think maybe just `sleep(i)` can work and is a bit easier on the eye?

guides/asynchronous-tasks/readme.md
ioquatix: Yes, the task will still enter the failed state though in any case where an exception propagates up to the root of the fiber running the task. I suppose your change is okay because it's just mentioning one case, but both cases are true - i.e.:

> (or the `failed` state if it raises an unhandled `StandardError`-derived exception or any exception)

But I think we can just say "raises an unhandled exception" since in every case the task enters the `failed` state.

ioquatix: (If you think it's confusing, I'm also okay with your change, but it should be noted that any exception that propagates up causes the task to enter a `failed` state.)

guides/asynchronous-tasks/readme.md
```diff
@@ -222,15 +224,15 @@ Async do
 end
 ```
 
-If you're letting individual tasks held by a barrier raise unhandled exceptions, be sure to call ({ruby Async::Barrier#stop}) to stop the remaining tasks:
+Unless your tasks all rescue and suppresses `StandardError`-derived exceptions, be sure to call ({ruby Async::Barrier#stop}) to stop the remaining tasks:
```
ioquatix, line 227: I think you can also suppress any exceptions and it's okay, even if it's questionable to do so.

guides/asynchronous-tasks/readme.md
```diff
@@ -264,7 +266,7 @@ As tasks run synchronously until they yield back to the reactor, you can guarant
 
 ## Exception Handling
 
-{ruby Async::Task} captures and logs exceptions. All unhandled exceptions will cause the enclosing task to enter the `:failed` state. Non-`StandardError` exceptions are re-raised immediately and will generally cause the reactor to fail. This ensures that exceptions will always be visible and cause the program to fail appropriately.
+{ruby Async::Task} captures and logs exceptions. All unhandled `StandardError`-derived exceptions will cause the enclosing task to enter the `:failed` state. Non-`StandardError` exceptions are re-raised from the reactor and thus cause it to fail. This ensures that exceptions will always be visible and cause the program to fail appropriately.
```
ioquatix, line 269:
```suggestion
{ruby Async::Task} captures and logs exceptions. All unhandled exceptions will cause the enclosing task to enter the `:failed` state. Non-`StandardError` exceptions are re-raised immediately and will cause the reactor to exit. This ensures that exceptions will always be visible and cause the program to fail appropriately.
```

I suppose we could go with this. The exceptions are not reprised from the reactor - they are reprised from the fiber which generally causes the reactor to exit if it was the one that caused resume. Internally, it's more complex.

ioquatix:
This looks good to me, I've given a few minor feedbacks. We should be clear that the failed state is entered by any kind of unhandled exception.

ioquatix:
Moving to https://github.com/socketry/async/pull/248 so I can address the feedback before merging.

Issue #237, opened 2023-06-02
Suggested API change


ColinDKelley:
## Problem Statement
Recently I've worked with a team of ~6 developers converting two services from `EventMachine` + `Synchrony` to `Async`. That has gone well, but one detail of the `Async` API has tripped us all up at first and in some cases, more than once.

The confusion arises from `Async do ... end` being overloaded:
1. At the outer edge, `Async do ... end` starts up the reactor that must surround the concurrent code.
2. Inside the reactor block, `Async do ... end` creates a concurrent `Async::Task` within the reactor.

When writing concurrent code (case 2), it is necessary that it be surrounded by a reactor. If that is missing, case 2 will be misconstrued as case 1, and it won't be concurrent at all. This is an easy mistake to make, especially when writing tests.

## Mitigation
To address this risk, we like follow Design by Contract, which in this case means that concurrent code (case 2) should ensure that there is a surrounding reactor before it proceeds. Here's how we've done that, crudely:

```ruby
def process_jobs(jobs)
  Async::Task.current? or raise ArgumentError, "must start up reactor first with surrounding 'Async do ... end'"

  jobs
    .map { Async { _1.process! } }
    .map(&:wait)
end
```

The wrapper code that creates the reactor looks like:

```ruby
Async do
  process_jobs(jobs)
end
```

## Proposal
An ideal solution would separate the overloaded cases so that the contract can be implicitly enforced by the `Async` gem without clients needing to write manual code each time.

Since case 2 appears much more frequently in code than case 1, let's leave case 2 as-is, but change case 1 to `Async.run do ... end`. So the wrapper for the above code would read:

```ruby
Async.run do
  process_jobs(jobs)
end
```

With this contract, case 2 could automatically raise an exception on this line if it is run without a surrounding reactor:
```
    .map { Async { _1.process! } }

=> Async::ReactorNotRunning: You should start a reactor first with `Async do ... end` surrounding this code.
```

## Migration
We obviously would not want to introduce this proposal as a breaking change. Instead we could phase it in:

### Phase 1
Introduce `Async.run` as a preferred but optional way to create the reactor at the outer level. Update all documentation to show it as the preferred way to create the reactor.

### Phase 2
Deprecate `Async do ... end` being called without a surrounding reactor. In the deprecation warning, indicate that it should be written as `Async.run do ... end` if it is the outer level that is creating the reactor.

### Phase 3
[...]

PR #238, opened 2023-06-02
Introduce some new `Schedule` operation for explicit nesting.

ioquatix:
This is an alternative to the proposal in <https://github.com/socketry/async/issues/237>.

It requires the following:

```ruby
Schedule {} # (1) Error, No parent async task set!

Sync do
  Schedule {...} # (2) Ok, running in an existing event loop / reactor.
end
```

I have no opinion about the name right now.

The original proposal has this example code:

```ruby
def process_jobs(jobs)
  Async::Task.current? or raise ArgumentError, "must start up reactor first with surrounding 'Async { ... }"

  jobs
    .map { Async { _1.process! } }
    .map(&:wait)
end
```

It would become this:

```ruby
def process_jobs(jobs)
  jobs
    .map { Schedule { _1.process! } }
    .map(&:wait)
end
```

However, I think it's worse, because the error of the missing event loop is delayed until some processing is already done. It also begs the question, why such an error is useful. If the expectation is that method can run concurrently, just make it so (see "Using Outer `Sync` Instead", below).

The alternative design avoids this:

```ruby
def process_jobs(jobs, parent: Async::Task.current) # Error will occur here if no parent async task is set.
  jobs
    .map { parent.async { _1.process! } }
    .map(&:wait)
end
```

It's also going to be a little less efficient to look up the parent context on every iteration, and it also gives you less control (e.g. providing an explicit semaphore to `parent:`).

By the way, it should be noted, that `Fiber.schedule` is very similar to the desired functionality:

```ruby
irb(main):001:0> Fiber.schedule{}
(irb):1:in `schedule': No scheduler is available! (RuntimeError)

vs

irb(main):003:0> Async{Fiber.schedule{}}
=> #<Async::Task:0x000000000000d5fc>
```

It should also be noted that my original proposal was `Fiber{}` rather than `Fiber.schedule{}`. In the original proposal, it could be used like this:

```ruby
def process_jobs(jobs)
  jobs
    .map { Fiber.schedule { _1.process! } }
    .map(&:wait)
end
```

## Using Outer `Sync` Instead

The point of the `Async` interface is to avoid the user having to be aware of the parent concurrency, however, I do accept there are a few sharp edge cases. For example, if you are expecting `Async` to run in conjunction with other tasks.

```ruby
def process_jobs(jobs)
  Sync do |task| # Ensures a reactor is running, and if not, starts it for this block.
    jobs
      .map { task.async { _1.process! } } # All of these tasks run concurrently now.
      .map(&:wait)
  end
end
```

For the purpose of this library, I'm not very enthusiastic about breaking backwards compatibility, but I'm okay to introduce new features if they help the code become more expressive and less error-prone.

## Contribution

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -234,7 +234,7 @@ module Async
 		# @returns [Task]
 		# @raises[RuntimeError] If task was not {set!} for the current fiber.
 		def self.current
-			Thread.current[:async_task] or raise RuntimeError, "No async task available!"
+			Thread.current[:async_task] or raise RuntimeError, "No parent async task set!"
 		end
 		
 		# Check if there is a task defined for the current fiber.
--- a/lib/kernel/schedule.rb
+++ b/lib/kernel/schedule.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2019-2022, by Samuel Williams.
+
+require_relative "../async/reactor"
+
+module Kernel
+	def Schedule(...)
+		::Async::Task.current.async(...)
+	end
+end


lib/async/task.rb
```diff
@@ -234,7 +234,7 @@ def stop(later = false)
 		# @returns [Task]
 		# @raises[RuntimeError] If task was not {set!} for the current fiber.
 		def self.current
-			Thread.current[:async_task] or raise RuntimeError, "No async task available!"
+			Thread.current[:async_task] or raise RuntimeError, "No parent async task set!"
```
ColinDKelley, line 237: This wording is definitely an improvement. 👍

PR #253, opened 2023-06-19
Guide: ensure barrier is in scope

olleolleolle:
This change avoids a NoMethodError on the barrier local variable. And also a LoadError, with the Async::Barrier constant.

I was following the guide, and when I ran my code, I found these issues, and this change fixed them.

## Types of Changes

- Bug fix.
- Maintenance.

## Contribution

- [x] I tested my changes locally.
- [x] I agree to the [Developer's Certificate of Origin 1.1](https://developercertificate.org/).

PR #254, opened 2023-06-19
Follow-up barrier in scope docs fix

olleolleolle:
This makes the same change as #253 but in this file: ensure block may not have access to the `barrier` local.

## Types of Changes

- Bug fix.
- Maintenance.

## Contribution


ioquatix:
Thanks!!

Issue #260, opened 2023-07-22
Unhandled Exception does not stop program running


jsaak:
Unhandled Exception does not stop program running:

(does work when uncommenting Fiber.schedule do)

```
require 'async'

Fiber.set_scheduler(Async::Scheduler.new)

class X
  def set_callback(&block)
    @block = block
  end

  def run_callback
    @block.call
  end

  def run
    Fiber.schedule do
      loop do
        puts 1
        sleep 1
      end
    end
  end
end

#Fiber.schedule do
  x = X.new
  x.set_callback do
    puts 3
    asdf
    puts 4
  end

  x.run
  sleep 1
  puts 2
  x.run_callback
#end

```

ioquatix:
This is probably a bug that can only be fixed by modifying CRuby. I'll take a look.

ioquatix:
As a bit of a hack, we can define `Async::Scheduler#scheduler_close` as:

```ruby
		def scheduler_close
			unless $!
				self.run
			end
		ensure
			self.close
		end
```

and it works as you expect.

ioquatix:
I'm basically okay with the proposed fix, the chance of `$!` being set incorrectly in the `scheduler_close` context is probably zero.

ioquatix:
This should be fixed, please feel free to test it and report back.

jsaak:
Either this fix or the scheduler closing broke exception handling for me:
I can not catch exception any more in Fibers.

```
/home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:246:in `select': Failed (RuntimeError)
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:246:in `run_once!'
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:222:in `run_once'
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:285:in `block in run'
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:279:in `handle_interrupt'
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:279:in `run'
	from /home/jsaak/.gem/ruby/3.2.2/gems/async-2.6.3/lib/async/scheduler.rb:46:in `scheduler_close'
[...]

PR #262, opened 2023-07-22
Fiber scheduler exception

ioquatix:
Fixes https://github.com/socketry/async/issues/260

## Types of Changes

- Bug fix.
- New feature.

## Contribution


ioquatix:
This is a bit of a hack: CRuby is not totally consistent here - threads behave a little differently than top level process schedulers. This should probably be fixed.

PR #266, opened 2023-08-09
add test for timeout while getting from stdin

emiltin:
Added a test to shows that with_timeout() does not return Async::Timeout() if timing out while getting from stdin.

The test can be run with:
```
% bundle exec sus test/async/task.rb:599
🏁 Finished in 103.3ms; 9.682 assertions per second.
🐇 No slow tests found! Well done!

🤔 Failed assertions:
describe Async::Task with #with_timeout it will timeout while getting from stdin test/async/task.rb:599
	expect #<Errno::ENOENT: No such file or directory @ io_fillbuf - fd:0 <STDIN>> to
		be is_a? Async::TimeoutError
			✗ assertion failed test/async/task.rb:612
```

## Types of Changes
- Added a test

## Contribution

### DIFF
--- a/test/async/task.rb
+++ b/test/async/task.rb
@@ -596,6 +596,22 @@ describe Async::Task do
 			expect(state).to be == :timeout
 		end
 		
+		it "will timeout while getting from stdin" do
+			error = nil
+			
+			reactor.async do |task|
+				begin
+					task.with_timeout(0.1) { STDIN.gets }
+				rescue StandardError => e
+					error = e
+				end
+			end
+			
+			reactor.run
+			
+			expect(error).to be_a(Async::TimeoutError)
+		end
+
 		it "won't timeout if execution completes in time" do
 			state = nil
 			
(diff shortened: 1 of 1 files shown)


test/async/task.rb
```diff
@@ -596,6 +596,22 @@ def after
...
 		end
 		
+		it "will timeout while getting from stdin" do
+			error = nil
+			
+			reactor.async do |task|
+				begin
+					task.with_timeout(0.1) { STDIN.gets }
+				rescue StandardError => e
+					error = e
```
ioquatix, line 606:
```suggestion
				rescue StandardError => error
				  # Ignore.
```

test/async/task.rb
```diff
@@ -596,6 +596,22 @@ def after
 			expect(state).to be == :timeout
 		end
 		
+		it "will timeout while getting from stdin" do
+			error = nil
+			
+			reactor.async do |task|
+				begin
+					task.with_timeout(0.1) { STDIN.gets }
```
ioquatix, line 604:
```suggestion
					task.with_timeout(0.1) {STDIN.gets}
```

ioquatix:
This test passes with `Select`, `URing` and fails on `EPoll`. Not sure why, investigating.

Issue #256, opened 2023-06-29
gets() hangs on Windows


emiltin:
Hi,
Running this on Mac and Linux:

```ruby
input, output = IO.pipe
Async do
  puts "starting"
  Async do
    puts "reading..."
    line = input.gets
    puts "got #{line.inspect}"
  end
  puts "writting..."
  output.puts "hello"
end
```

outputs:
```
starting
reading...
writting...
got "hello\n"
```

But on Windows it just outputs:
```
starting
reading...
```

and then hangs. It seems the call to `input.gets()`blocks?

I'm running the Windows code in a Github Workflow:
Microsoft Windows Server 2022, 10.0.20348
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x64-mingw-ucrt]
Using async 2.6.2
Using async-io 1.35.0
Using bundler 2.4.13

emiltin:
fixed the text, on Windows this is output:
```
starting
reading...
```

emiltin:
I cannot reproduce this with async 2.6.3, closing.

ioquatix:
Wow great, thanks for reporting back!

emiltin:
Hi,
Any plans for adding Windows to the test matrix? This seems still to hang on Windows.

ioquatix:
Yes, but it's slow. I started writing support for IOCP in `io-event` gem.

emiltin:
Cool, thank you for the response.

PR #267, opened 2023-08-10
test gets in task

emiltin:
Adds a test for #256

Test can be run with:
```
% bundle exec sus test/async/task.rb:812                             
1 passed out of 1 total (1 assertions)
🏁 Finished in 723.0µs; 1383.126 assertions per second.
🐇 No slow tests found! Well done!
```

It succeeds on Mac, but I suspect it will hang on Windows, as it's essentially the example from #256.

## Types of Changes
- added a test

## Contribution
- [x] I tested my changes locally. (not on Windows though..)

### DIFF
--- a/test/async/task.rb
+++ b/test/async/task.rb
@@ -836,4 +836,16 @@ describe Async::Task do
 			end
 		end
 	end
+
+	it "can gets in a task" do
+		IO.pipe do |input, output|
+		  Async do
+				Async do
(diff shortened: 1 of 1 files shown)


emiltin (commented):
ok

ioquatix:
We should probably try to add Windows to the CI matrix.

ioquatix (approved):

PR #271, opened 2023-08-23
Some basic memory usage benchmarks.

ioquatix:
Compare memory usage of threads vs fibers.

cc @MatheusRich

## Contribution

### DIFF
--- a/benchmark/memory/fibers.rb
+++ b/benchmark/memory/fibers.rb
@@ -1,0 +1,18 @@
+#!/usr/bin/env ruby
+
+pid = fork do
+	fibers = 100.times.map do
+		Fiber.new{loop{Fiber.yield}}.resume
+	end
+	
+	sleep
+end
+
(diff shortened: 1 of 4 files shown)


benchmark/memory/fibers.rb
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+
+pid = fork do
+	fibers = 100.times.map do
+		Fiber.new{loop{Fiber.yield}}.resume
```
MatheusRich, line 5: I might have missed something, but doesn't this only run once, while the thread example sleeps forever?

MatheusRich:
Thank you!

benchmark/memory/fibers.rb
ioquatix: All the fibers and threads are still alive at the point where you measure memory usage. Does that seem sufficient?

PR #276, opened 2023-08-30
Allow queue to be used for task `finished`.

ioquatix:
```ruby
#!/usr/bin/env ruby

require_relative 'lib/async'
require_relative 'lib/async/queue'

Async do
  queue = Async::Queue.new
  
  10.times do |i|
    Async(finished: queue) do
      # raise "Boom" if rand > 0.5
      sleep(i)
    end
  end
  
  queue.each do |task|
    puts task.wait
  end
end
```

However, it's not clear how to close the queue to indicate no more jobs will be added.


emiltin:
I'm sure I understand how this would be used to make a barrier fail fast in case a task fails.

Is the idea that all tasks in the barrier use the same condition to signal when they are done, so that the barrier can used this to fail fast? I think that would have the side effect that if some other code wait a particular task in the barrier, then that wait would end whenever any task in the barrier finishes?

emiltin:
> However, it's not clear how to close the queue to indicate no more jobs will be added.

What about a `drain` method, which works like `each`, except it returns as soon as the queue is empty?

emiltin:
> What about a `drain` method, which works like `each`, except it returns as soon as the queue is empty?

To clarify - this would imply that tasks are added to the queue when created, and are removed when they finish. `drain` would return each task as they complete, returning when there is no more items.

The current PR works differently: it adds task to the queue only when finished. `each` waits for new items to be added, which are then removed and returned. It never returns.

ioquatix:
I have not had time to revisit this yet, but I will in the coming week.

PR #313, opened 2024-04-04
Update readme.md - fixing a typo

peychinov:
Just a typo fix.

### DIFF
--- a/guides/getting-started/readme.md
+++ b/guides/getting-started/readme.md
@@ -1,6 +1,6 @@
 # Getting Started
 
-This guide gives shows how to add async to your project and run code asynchronously.
+This guide shows how to add async to your project and run code asynchronously.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks!

To fix this properly, please fix the source document: https://github.com/socketry/async/tree/main/guides/getting-started

Then run:

```
bundle exec bake utopia:project:readme:update
```

Then submit both files in this PR. Thanks!

olleolleolle:
@peychinov Perhaps, `group :maintenance, optional: true do` the gems that have these recipes, perhaps they're not included in your current installation?

So, to get those:

```
bundle config with maintenance
bundle
```

ioquatix (approved):
Thanks!

ioquatix:
@olleolleolle is correct - in the past, I've had problems installing gems which I use for release management, documentation, and general maintenance in CI infrastructure, so I decided to split it out into an optional group. Sorry it caused confusion and thanks for giving it a go!

PR #326, opened 2024-07-11
 Node#parent: fix documentation

paddor:


ioquatix:
Thanks for this.

However, after checking, the syntax of the subsequent attribute is wrong, not this one:

https://github.com/ioquatix/decode/blob/9d6f88e84116d999ddd77eb10a68c2c3417535ba/lib/decode/comment/attribute.rb#L15

The "name" is not a required feature of the attribute tag. So I'll remove the subsequent one and close this PR without merging.

Thanks for your contribution.

paddor:
I don't see what's wrong with the subsequent `#children` attribute. But the documentation of the `#parent` attribute currently looks [like this](https://www.rubydoc.info/gems/async/Async/Node#The+parent+node%2E=-instance_method):

```
#The parent node.=(parentnode. = (value)) ⇒ Object
```

The missing name is the only see I noticed is different from the other attributes.

ioquatix:
That looks messed up, let me check it.

ioquatix:
At least according to YARD, I don't think the attribute name is required, especially if it's just repeating what the attribute is already called. Maybe it's something wrong with rubydoc.info? What documentation generation tool are they using?

paddor:
Yeah, it should be YARD. It doesn't really now the `@attribute` directive. It has `@!attribute`, which is for attributes created with meta-programming.

I usually document normal attributes like this:
```
      # @return [Messaging::Handlers::Base]
      attr_reader :messaging
```

It's not very intuitive. Oh well.

Utopia looks very nice but what I miss are superclasses (https://github.com/socketry/utopia-project/issues/3), inherited methods, and private methods. It just makes a quick reading of a method's implementation that much more comprehensible.

ioquatix:
I think we can add those things - let me take a look.

PR #336, opened 2024-08-04
Graceful shutdown

ioquatix:
In general, users should expect ensure blocks to work normally (sequentially).

Transient tasks do not keep the reactor alive. e.g.

```ruby
Async do
  Async(transient: true) do
    sleep
  ensure
    sleep 1; $stderr.write "."
  end
end
```

Before this PR, the above program will not print any dot. After this PR, the program will print a dot. This allows transient tasks to clean up more robustly. The downside is, a badly behaving transient task may stall the event loop from exiting. However, `Ctrl-C` will interrupt the event loop again and cause it to exit.

## Types of Changes

- Bug fix.
- New feature.

## Contribution

### DIFF
--- a/lib/async/idler.rb
+++ b/lib/async/idler.rb
@@ -41,7 +41,8 @@ module Async
 			backoff = nil
 			
 			while true
-				load = scheduler.load 
+				load = scheduler.load
+				
 				break if load < @maximum_load
 				
 				if backoff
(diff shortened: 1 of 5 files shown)


ioquatix:
```ruby
require_relative 'lib/async'

reactor = nil

begin
	Async do |task|
		reactor = task.reactor
		
		Async(transient: true) do
			$stderr.puts "Hello, World!"
			sleep
		ensure
			$stderr.puts "Goodbye, World!"
			sleep
		end
	end
ensure
	binding.irb
end
```

Running this program and pressing Ctrl-C confirms the behaviour too.

PR #348, opened 2024-10-01
Kernel#Sync: pass annotation: keyword to newly created Reactor

paddor:
Just for symmetry with `Kernel#Async`. In a library I changed the specs to always run in a `Sync {...}` block which meant I had to change the main reactor from `Async(annotation: 'x') { ... }` to `Async(annotation: 'x') { ... }.wait` because `Kernel#Sync` didn't take an annotation. This PR allows it to be just `Sync(annotation: 'x') { ... }`.

### DIFF
--- a/lib/kernel/sync.rb
+++ b/lib/kernel/sync.rb
@@ -17,7 +17,11 @@ module Kernel
 	# @asynchronous Will block until given block completes executing.
 	def Sync(*arguments, **options, &block)
 		if task = ::Async::Task.current?
-			yield task
+			if annotation = options[:annotation]
+				task.annotate(annotation) { yield task }
+			else
+				yield task
+			end
 		elsif scheduler = Fiber.scheduler
 			::Async::Task.run(scheduler, &block).wait
 		else
--- a/lib/kernel/sync.rb
+++ b/lib/kernel/sync.rb
@@ -18,7 +18,7 @@ module Kernel
 	def Sync(annotation: nil, &block)
 		if task = ::Async::Task.current?
 			if annotation
-				task.annotate(annotation) { yield task }
+				task.annotate(annotation) {yield task}
 			else
 				yield task
 			end
--- a/lib/kernel/sync.rb
+++ b/lib/kernel/sync.rb
@@ -15,9 +15,9 @@ module Kernel
 	#
 	# @public Since `stable-v1`.
 	# @asynchronous Will block until given block completes executing.
-	def Sync(*arguments, **options, &block)
+	def Sync(annotation: nil, &block)
 		if task = ::Async::Task.current?
-			if annotation = options[:annotation]
+			if annotation
 				task.annotate(annotation) { yield task }
 			else
 				yield task
@@ -29,7 +29,13 @@ module Kernel
 			reactor = Async::Reactor.new
 			
 			begin
-				return reactor.run(*arguments, **options, finished: ::Async::Condition.new, &block).wait
+				options = {
+					finished: ::Async::Condition.new
+				}
+
+				options[:annotation] = annotation if annotation
+
+				return reactor.run(**options, &block).wait
 			ensure
 				Fiber.set_scheduler(nil)
 			end
--- a/lib/kernel/sync.rb
+++ b/lib/kernel/sync.rb
@@ -29,13 +29,7 @@ module Kernel
 			reactor = Async::Reactor.new
 			
 			begin
-				options = {
-					finished: ::Async::Condition.new
-				}
-
-				options[:annotation] = annotation if annotation
-
-				return reactor.run(**options, &block).wait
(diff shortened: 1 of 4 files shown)


ioquatix:
How do you think we should handle the following case?

```ruby
Async(annotation: "foo") do
  Sync(annotation: "bar") do |task|
    puts task.annotation
  end
end
```

paddor:
Could `Kernel#Sync` replace the annotation of the current task and then restore it to the previous value?

ioquatix:
> Could Kernel#Sync replace the annotation of the current task and then restore it to the previous value?

Yeah, I considered that too. I think it's a reasonable idea.

paddor:
`Task#annotate` / `Fiber#annotate` would be nice but they don't yield the task to the block.

paddor:
Ah wait, that's not a big deal. I'm working on it...

ioquatix:
We can always update the fiber annotation gem to support `...` arguments to yield or something like that.

lib/kernel/sync.rb
```diff
@@ -25,7 +25,7 @@ def Sync(&block)
 			reactor = Async::Reactor.new
 			
 			begin
-				return reactor.run(finished: ::Async::Condition.new, &block).wait
+				return reactor.run(*arguments, **options, finished: ::Async::Condition.new, &block).wait
```
ioquatix, line 28: I don't think we should pass through all options as in some cases they won't be used... it would be confusing if some cases they have an effect and other cases they don't.

paddor: I've addressed that in c166dcc

paddor:
I think this is ready.

lib/kernel/sync.rb
```diff
@@ -29,7 +29,13 @@ def Sync(*arguments, **options, &block)
...
 			
 			begin
-				return reactor.run(*arguments, **options, finished: ::Async::Condition.new, &block).wait
+				options = {
+					finished: ::Async::Condition.new
+				}
+
+				options[:annotation] = annotation if annotation
+
+				return reactor.run(**options, &block).wait
```
ioquatix, line 38: Can't you just pass `annotation: annotation`?

paddor: No, then some tests won't pass.

paddor: I tried again. The tests do work now with the simpler variant.

lib/kernel/sync.rb
```diff
@@ -15,17 +15,21 @@ module Kernel
 	#
 	# @public Since `stable-v1`.
 	# @asynchronous Will block until given block completes executing.
-	def Sync(&block)
+	def Sync(annotation: nil, &block)
 		if task = ::Async::Task.current?
-			yield task
+			if annotation
+				task.annotate(annotation) { yield task }
```
ioquatix, line 21:
```suggestion
				task.annotate(annotation) {yield task}
```

ioquatix (approved):

ioquatix:
Thanks for your contribution.

I understand the value of this change, and think that it's ultimately a good change.

However, I want to point out there will be a slight performance hit. Hopefully things like YJIT mitigate it on the fast path (no annotation, existing task).

I also had some thoughts around how to present annotations in backtrace, e.g. an annotation is stack frame specific and gets added to exception messages, etc. In other words, there is more work to be done in this area FYI.

paddor:
Thanks for the guidance and merging.

I'd be a little surprised if the performance hit was even measurable. 

Annotations in the backtrace sounds like a good idea. Let me know if I can help in any way.

ioquatix:
Annotating the call stack would be an alternative to annotating the fiber. However, for the purpose of Async, I think it wouldn't look that different.

We'd need to write up a proposal on <bugs.ruby-lang.org> and propose a similar but non-finer specific interface for "annotate the current stack frame". If it was accepted, then `Sync(annotation:)` would simply use that.

ioquatix:
https://bugs.ruby-lang.org/issues/19056 was the previous discussion, and at the meeting we discussed having per-call-stack annotations (could be added to exception backtrace too).

paddor:
So something like `Callstack.annotate(my_annotation) { ... }`? And I guess (because Matz said it would have to be stacked) it would have to append the annotation into an array (like breadcrumbs), so `Callstack.annotations` would return the entire Array (like `Class#ancestors`), and pop from that array when the annotated block is done.

What did you mean by "APM" in that discussion on bugs.ruby-lang.org? If it's advanced power management, I don't know how that relates to this.

ioquatix:
> Application performance monitoring (APM) is the practice of tracking key software application performance metrics using monitoring software and telemetry data. Practitioners use APM to ensure system availability, optimize service performance and response times, and improve user experiences.

PR #352, opened 2024-11-06
Add support for `Fiber::Scheduler#blocking_operation_wait`.

ioquatix:
This is a very basic implementation of the `Fiber::Scheduler#blocking_operation_wait` hook.

See <https://bugs.ruby-lang.org/issues/20876> for more context.

## Types of Changes

- New feature.
 
## Contribution

### DIFF
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -338,6 +338,25 @@ module Async
 			return @selector.process_wait(Fiber.current, pid, flags)
 		end
 		
+		# Wait for the given work to be executed.
+		#
+		# @public Since *Async v2.19* and *Ruby v3.4*.
+		# @asynchronous May be non-blocking.
+		#
+		# @parameter work [Proc] The work to execute on a background thread.
+		# @returns [Object] The result of the work.
+		def blocking_operation_wait(work)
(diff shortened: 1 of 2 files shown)


lib/async/scheduler.rb
```diff
@@ -291,6 +291,16 @@ def process_wait(pid, flags)
 			return @selector.process_wait(Fiber.current, pid, flags)
 		end
 		
+		def blocking_operation_wait(work)
```
olleolleolle, line 294: What if this section had an API documentation note, telling the reader what Ruby version this exists in, etc?

```suggestion
        # Implements Fiber::Scheduler#blocking_operation_wait support. This exists in Ruby 3.3.x.
		def blocking_operation_wait(work)
```

olleolleolle: My version number there is just a placeholder for the real thing.

ioquatix: Ah yes, this should land in 3.4. It's probably a good idea to document more of the scheduler.

ioquatix:
Confirmed tests passing on ruby-head.

tenderlove:
Does this mean that every time something releases the GVL, async will create a new thread?

PR #367, opened 2025-01-03
Redefine aliases for LimitedQueue

jahfer:
👋🏼 I stumbled into this bugfix by only partially reading the documentation 😅 

I was playing with `Async::LimitedQueue`, but called `#pop` on it (which is technically _not_ documented), and it completely hung. Reason being that the `@full` notification doesn't fire on `#pop` because it's referring to the implementation on `Async::Queue` rather than using `Async::LimitedQueue#dequeue`, which correctly signals `@full`.

As a fix, I redefined the aliases in `Async::LimitedQueue` so they pick up the redefinitions. Without the changes, the updated tests will completely hang in the `#pop` case. 

I also noticed the original test for this wasn't actually working because `#contain_exactly` wasn't defined, and so the async block containing the assertions was being completely ignored. I couldn't get the structure of what was there prior to work (even with switching to `#have_value`), so I asserted based on the dequeued values instead. No strong opinion on this :)

## Types of Changes

- Bug fix.

## Contribution

### DIFF
--- a/lib/async/queue.rb
+++ b/lib/async/queue.rb
@@ -125,13 +125,16 @@ module Async
 		# If the queue is full, this method will block until there is space available.
 		#
 		# @parameter item [Object] The item to add to the queue.
-		def <<(item)
+		def push(item)
 			while limited?
 				@full.wait
 			end
 			
 			super
 		end
+
+		# Compatibility with {::Queue#push}.
+		alias << push
 		
 		# Add multiple items to the queue.
 		#
@@ -163,5 +166,8 @@ module Async
 			
 			return item
 		end
+		
+		# Compatibility with {::Queue#pop}.
+		alias pop dequeue
 	end
 end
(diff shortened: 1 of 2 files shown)


ioquatix:
TBH, aliases and sub-classes are kind of confusing in Ruby leading to behaviour like this. Thanks for the PR, I think it looks okay, but I might replace the aliases with direct dispatch (e.g. `def <<(...); self.push(...); end`)

ioquatix (approved):

ioquatix:
FYI, I'm slowly leaning away from using `Async::Queue` and so on. During Async v1, they were useful, but in Async v2, I believe it is just better to use `Thread::Queue` and so on. Althought, while I have not tested it, there may be a small performance advantage to using `Async::Queue` etc.

ioquatix:
Okay, I did this: https://github.com/socketry/async/commit/a91f32c10d3eb14980f159326a05aceefd6ba272

PR #371, opened 2025-01-27
add Async::Promise

### LINKED ISSUE #370: Async::Variable#reject or maybe Async::Promise
I used a `Async::Variable` to get a value back from an asynchronous call to a C lib (Open62541). That works in the happy path. The C lib actually makes the callback with a status (good/bad) and the value. If the status is bad, there is no value. In that case, I'd want `Async::Variable#wait` to raise an exception.

```ruby
class Client
  def read_namespace_array
    var = Async::Variable.new

    read_value NodeIds::Variables::SERVER_NAMESPACE_ARRAY do |status, value|
      # TODO: check status
      var.resolve value
    end

    var.value
  end
end
```

Simply resolving the `Async::Variable` with an exception does not work.

```ruby
require 'async'
require 'async/variable'


Sync do |t|
  var = Async::Variable.new

  t.async do
    p var.wait
  ensure
    p exception: $!
  end

  t.async do
    var.resolve RuntimeError.new
  end
end
```

Output:
```
#<RuntimeError: RuntimeError>
{exception: nil}
```

Would it be smart to add `Async::Variable#reject(exception)`? Instead of setting `@value`, it would set `@error`, which would be raised by `#wait`. I guess `#resolved?` should continue to return `false` in that case, and a new method `#rejected?` would return `true`.

To avoid modifying the existing class, maybe an `Async::Promise < Async::Variable` could be added. What do you think about this?

paddor:
This is a draft to fix #370

### DIFF
--- a/lib/async/promise.rb
+++ b/lib/async/promise.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2025, by Patrik Wenger.
+
+require_relative "variable"
+
+module Async
+
+	# Similar to a Variable, in that it allows a task to wait for a value to resolve, but also supports rejection given
+	# an error. The given error is raised as an exception in the waiting task.
+	class Promise < Variable
+		def initialize(...)
+			super
+
+			@error = nil
+		end
+
+		def reject(error = RuntimeError.new('promise rejected'), value = nil)
+			@error = error
+			resolve value
+		end
+
+		def rejected?
+			resolved? && !!@error
+		end
+
+		# @returns [Boolean] Whether the value has been resolved.
+		# @raises [Exception] The error with which this Variable was rejected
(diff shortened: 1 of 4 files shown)


ioquatix:
- Missing documentation for `reject` and `rejected?`.

On the surface, I'm okay with this change. However, I am not sure I want to merge it yet.

I wonder if `resolved?` should return true if the variable was rejected. At least, that feels more natural to me e.g. from a usage point of view. The point of such a method is to control level signalling, e.g. https://github.com/socketry/async-http/blob/e2598452c5d014623cee16b451f1cb84e4e74f27/lib/async/http/protocol/http1/finishable.rb#L18

I wonder if we should encourage users just to use `Async` in the case they need a full-blown promise style interface.

Maybe the code above can be simplified with this change.

I would like to explore this further.

PR #380, opened 2025-03-26
Correct comment about when blocking_operation_wait() was released

XrXr:
The hook is not in the v2.19.0 tag.
https://github.com/socketry/async/commit/c317990a4f44f14d43c27d77d663abfcfab89b62 shows that it was first added with v2.21.0.

### DIFF
--- a/lib/async/worker_pool.rb
+++ b/lib/async/worker_pool.rb
@@ -14,7 +14,7 @@ module Async
 		module BlockingOperationWait
 			# Wait for the given work to be executed.
 			#
-			# @public Since *Async v2.19* and *Ruby v3.4*.
+			# @public Since *Async v2.21* and *Ruby v3.4*.
 			# @asynchronous May be non-blocking.
 			#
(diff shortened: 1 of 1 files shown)


ioquatix (approved):

ioquatix:
Thanks for correcting this!

ioquatix:
FYI, this feature is highly experimental, I believe we need a better implementation of the worker pool - probably a C extension so we avoid re-acquiring the GVL in a thread.

XrXr:
If this is highly experimental, maybe it should be opt-in instead of opt-out (by not upgrading)? There's been 2 class of crashes related to this hook:

- https://bugs.ruby-lang.org/issues/21198
- https://bugs.ruby-lang.org/issues/21180

Moving a `nogvl` off to a different thread is inherently tricky to get right.

ioquatix:
Thanks for your suggestion and the links to the issues. 

I was under the impression that it is already opt-in: https://github.com/socketry/async/blob/main/lib/async/scheduler.rb#L28-L30

PR #384, opened 2025-04-24
Barrier waits in order of completion/failure.

ioquatix:
It can be confusing that a barrier waits on tasks in order rather than in the order of completion/failure, especially if a single failure implies the complete failure of the operation. This change ensures `Barrier#wait` returns tasks in order of completion.

## Types of Changes

- New feature.

## Contribution

- [ ] I agree to the [Developer's Certificate of Origin 1.1](https://developercertificate.org/).


samuel-williams-shopify:
This was merged in 9e11aefd013b5df1d7acef6ab6ea2dbd6219e3a1

PR #389, opened 2025-05-20
Add configurable custom exception handlers for unhandled errors

moznion:
This change allows users to customize how Async handles unhandled exceptions—for example, by logging them to a different destination, adding custom metadata, or forwarding them to an error monitoring service.

In my case, I’m trying to introduce this library into our production services. Currently, one of the main hurdles is the difficulty of monitoring unhandled exceptions. The motivation behind this change is to enable sending those exceptions to the error tracker like Sentry.

## Types of Changes

- New feature.

## Contribution

### DIFF
--- a/lib/async/task.rb
+++ b/lib/async/task.rb
@@ -61,7 +61,7 @@ module Async
 				super
 			end
 		end
-		
+
 		# @deprecated With no replacement.
 		def self.yield
 			Fiber.scheduler.transfer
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks for your contribution and sorry for not getting back to you sooner.

I'm basically against adding all the extra complexity.

However, I understand your problem and there is a hook that should allow you to achieve what you want with minimum effort.

https://github.com/socketry/async/blob/ec3c8705cf519a3482e9504f34787c026a208ada/lib/async/task.rb#L382-L384

There are two options:

- Write your own `Async::Console` interface, the basic requirements are shown here: https://github.com/socketry/async/blob/ec3c8705cf519a3482e9504f34787c026a208ada/lib/async/console.rb
- Replace `Async::Task#warn` with your own implementation.

Ruby is already a dynamic language, so we don't need to introduce even more dynamic dispatch. The above two interfaces are official extension points, although replacing `#warn` is probably a little more kludgy.

Is that sufficient for your needs?

PR #406, opened 2025-07-27
Restrict async to MRI Ruby only

seuros:
## Summary

Remove JRuby and TruffleRuby from CI configuration, add required_ruby_engine restriction to gemspec, and simplify Ruby engine checks in scheduler.rb.

Async depends on gems that require C compilation (io-event), and Java engines already have their own async counterparts.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,12 +29,6 @@ jobs:
         
         include:
           - os: ubuntu
-            ruby: truffleruby
-            experimental: true
-          - os: ubuntu
-            ruby: jruby
-            experimental: true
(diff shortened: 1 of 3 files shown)


ioquatix:
The plan is to support them, it's just low priority.

https://github.com/socketry/io-event/pull/148
https://github.com/rubygems/rubygems/pull/8879

seuros:
Jruby has it's own async Implementation.  So anyone on JRuby already has primitives like Promises, Futures, and ThreadPoolExecutor, which don't require a fiber scheduler shim.

People using these 2 engines mix Java and Ruby in the same codebase.

PR #412, opened 2025-08-28
Add `Async::PriorityQueue` for Consumer-Priority Resource Allocation.

samuel-williams-shopify:
This PR introduces `Async::PriorityQueue`, a unique concurrency primitive that implements the **inverse of traditional task prioritization**. Instead of prioritizing work items, it prioritizes the **consumers** requesting work, enabling sophisticated resource allocation strategies based on who is asking rather than what is being processed.

- **Worker Class Hierarchies**: Critical system processes bypass background job consumers.
- **Load Balancing**: High-priority services get first access to shared resource pools.
- **Capacity Management**: Important consumers reserve access to constrained resources.

### Example

```ruby
Sync do
	queue = Async::PriorityQueue.new
	
	# Start consumers with different priorities:
	Async {puts queue.dequeue(priority: 1)}   # Low priority
	Async {puts queue.dequeue(priority: 10)}  # High priority
	
	# High priority consumer gets this first:
	queue.push("item")
end
```
## Types of Changes

- New feature.

## Contribution

### DIFF
--- a/fixtures/async/a_priority_queue.rb
+++ b/fixtures/async/a_priority_queue.rb
@@ -1,0 +1,195 @@
+# frozen_string_literal: true
+
+# Released under the MIT License.
+# Copyright, 2025, by Samuel Williams.
+# Copyright, 2025, by Shopify Inc.
+
+require "async"
+require "async/priority_queue"
+require "sus/fixtures/async"
+
+require "async/chainable_async"
+
+module Async
+	APriorityQueue = Sus::Shared("a priority queue") do
+		let(:queue) {subject.new}
+		
+		# Include all basic queue behaviors
+		it_behaves_like Async::AQueue
+		
+		with "priority ordering" do
+			it "serves consumers in priority order" do
+				results = []
+				
+				# Create consumers with different priorities
+				low = reactor.async do
+					results << queue.dequeue(priority: 1)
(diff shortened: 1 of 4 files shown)


franciscojmc:
## Consideration: Handling Dead Fibers in the Waiter Queue
Great implementation. I wanted to raise a potential edge case for discussion regarding fiber lifecycle management.

### Scenario to consider:

What happens if a fiber dies unexpectedly while waiting in the queue? Currently, the dead fiber's Waiter would remain in the priority heap, which could lead to items being lost:

```ruby
queue = Async::PriorityQueue.new

# Fiber starts waiting
fiber = Async do
  queue.dequeue(priority: 10)
end

# Fiber terminates unexpectedly
fiber.stop

# The Waiter object remains in @waiting heap

# Later, when we push an item:
queue.push("item1")
# This would:
# 1. Pop the dead fiber's waiter from heap
# 2. Assign "item1" to the dead waiter
# 3. Signal a condition variable no one is waiting on
# Result: "item1" is effectively lost
```

What are your thoughts on whether this should be handled?

ioquatix:
This implementation is both fiber and thread safe.

ioquatix (approved):
LGTM.

PR #441, opened 2026-01-22
Introduce `Task.wait_all`.

samuel-williams-shopify:
See https://github.com/socketry/async/pull/89 for previous discussion.

This provides a slightly more ergonomic interface for `task.children&.each(:wait)` which is a relatively common pattern.

## Types of Changes

- New feature.

## Contribution

### DIFF
--- a/lib/async/node.rb
+++ b/lib/async/node.rb
@@ -297,8 +297,6 @@ module Async
 		
 		# Wait for this node to complete. By default, nodes cannot be waited on.
 		# Subclasses like Task override this method to provide waiting functionality.
-		#
-		# @returns [self] Returns self for method chaining.
 		def wait
 			nil
 		end


lib/async/node.rb
```diff
@@ -295,6 +295,14 @@ def stop(later = false)
 			end
 		end
 		
+		# Wait for this node to complete. By default, nodes cannot be waited on.
+		# Subclasses like Task override this method to provide waiting functionality.
+		#
+		# @returns [self] Returns self for method chaining.
+		def wait
+			self
```
ioquatix:
```suggestion
			nil
```

lib/async/node.rb
```diff
@@ -295,6 +295,14 @@ def stop(later = false)
 			end
 		end
 		
+		# Wait for this node to complete. By default, nodes cannot be waited on.
+		# Subclasses like Task override this method to provide waiting functionality.
+		#
+		# @returns [self] Returns self for method chaining.
```
ioquatix:
```suggestion
```

Issue #455, opened 2026-06-18
Add Async::Condition#waiting_count


rmosolgo:
Hi! I've been expanding my use of Async in GraphQL-Ruby's `AsyncDataloader` and I have a use case for checking how many tasks are waiting for a given Async::Condition. 

What do you think of adding `Async::Condition#waiting_count`, like `Async::Queue#waiting_count`? Under the hood, it would use the same Ruby `Thread::Queue#num_waiting` API, for example: 

```ruby
def waiting_count
  @ready.num_waiting
end 
```

If you like it, I'd be happy to open a PR with this method and some tests. Thanks!

ioquatix:
It seems reasonable to me.

PR #456, opened 2026-06-22
Implement Condition#waiting_count

rmosolgo:
This adds `Condition#waiting_count` so you can tell how many tasks are waiting for a certain condition, as discussed in #455 

Please let me know if I can improve this PR in any way!

## Types of Changes

- New feature.

## Contribution

### DIFF
--- a/lib/async/condition.rb
+++ b/lib/async/condition.rb
@@ -32,7 +32,7 @@ module Async
 			!self.empty?
 		end
 		
-		# @returns [Integer] Number of fibers waiting on this condition
+		# @returns [Integer] Number of fibers waiting on this condition.
 		def waiting_count
 			@ready.num_waiting
 		end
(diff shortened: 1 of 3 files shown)


lib/async/condition.rb
```diff
@@ -32,6 +32,11 @@ def waiting?
 			!self.empty?
 		end
 		
+		# @returns [Integer] Number of fibers waiting on this condition
```
ioquatix:
```suggestion
		# @returns [Integer] Number of fibers waiting on this condition.
```

ioquatix (approved):

rmosolgo:
Thanks for taking a look and making that docs correction, @ioquatix. I see some test failures, but I don't think they're related to my changes here. Is there anything else I should do on this one?

ioquatix:
I had to fix the root cause. Released in v2.40.0. Note that checking counts can cause race conditions.

PR #459, opened 2026-06-29
Install default SIGINT handler

samuel-williams-shopify:
## Summary

- Add a small `Async::SIGINT` compatibility shim loaded by `require "async"`.
- Install the shim only on affected CRuby versions and only when SIGINT is still using the default handler.
- Add subprocess coverage for version gating, preserving existing handlers, and deferring SIGINT with `Thread.handle_interrupt`.

## Testing

- `bundle exec sus test/async.rb --verbose`
- `bundle exec rubocop -A lib/async.rb lib/async/sigint.rb test/async.rb`
- `bundle exec bake decode:index:coverage lib`
- `bundle exec sus`


samuel-williams-shopify:
I think fixing this upstream is a better idea.

PR #461, opened 2026-07-08
Propagate cancellation causes through task trees

infiton:
Propagate cancellation causes through `Node`, `Task`, and `Scheduler` cancellation traversal so bulk cancellation reuses the originating `Async::Cancel::Cause` instead of regenerating a fresh cause/backtrace for each descendant task.

This reduces redundant `Cancel::Cause.for` / backtrace capture work during scheduler-driven bulk cancellation and shutdown. Descendant tasks now observe the originating cancellation cause rather than a per-node synthetic cause generated during traversal.

A benchmark was added in `benchmark/cancel_cause.rb` to demonstrate the improvement against `main`:

  ```text
  # TASKS=1365
  main:   cause_for_calls=1365 allocated_objects=26009 duration_ms=7.42
  branch: cause_for_calls=1    allocated_objects=5545  duration_ms=5.79

  # TASKS=5461
  main:   cause_for_calls=5461 allocated_objects=103833 duration_ms=35.03
  branch: cause_for_calls=1    allocated_objects=21929  duration_ms=23.94
```

## Types of Changes

- Performance improvement.

## Contribution

### DIFF
--- a/lib/async/node.rb
+++ b/lib/async/node.rb
@@ -283,9 +283,10 @@ module Async
 		# Attempt to cancel the current node immediately, including all non-transient children. Invokes {#stop_children} to cancel all children.
 		#
 		# @parameter later [Boolean] Whether to defer cancelling until some point in the future.
-		def cancel(later = false)
+		# @parameter cause [Exception | Nil] The cause of the cancel operation.
+		def cancel(later = false, cause: $!)
 			# The implementation of this method may defer calling `stop_children`.
-			stop_children(later)
+			stop_children(later, cause: cause)
 		end
 		
 		# Backward compatibility alias for {#cancel}.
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: lib/async/node.rb [hunk 1]
```diff
@@ -295,9 +296,9 @@ def stop(...)
 		end
 		
 		# Attempt to stop all non-transient children.
-		private def stop_children(later = false)
+		private def stop_children(later = false, cause: $!)
```
tavianator: I don't think the new param needs a default value, it has only one caller (above) which already passes in the cause

infiton: 👍

ioquatix:
Thanks for your contribution. The only change I made was the state transition methods (ending with bangs) take positional arguments to be consistent with other state transition methods.

FYI, from my point of view, it's good to do less allocations during cancellation so this is a good idea. In general, error handling (which usually involves cancellation) should avoid doing unnecessary work. Otherwise you can have cascading failures (e.g. run out of memory).

PR #469, opened 2026-08-22
Retry interrupted waits in io_wait and process_wait

### LINKED ISSUE #467: Sleep/IO timeout can return early under concurrent per-fiber timeouts (async 2.42.0)

tavianator:
Fixes: https://github.com/socketry/async/issues/467

## Summary

- Retry `io_wait` after a stale selector wake-up while preserving the original timeout.
- Retry `process_wait` when the selector returns `false`, indicating that the native wait was interrupted before process completion.
- Preserve `nil` from `process_wait`, including legitimate `Process::WNOHANG` results.
- Require io-event `~> 1.21`, which provides consistent interrupted-process-wait behavior across native backends.
- Add backend-independent regression coverage for Async's retry policy.

## Scope

io-event owns the native interruption and cancellation contract; Async owns the higher-level retry policy. socketry/io-event#215 fixed modern io_uring cancellation and was released in io-event v1.21.0.

The separate threaded process-wait behavior is not worked around here. socketry/io-event#212 adds test-only coverage on Ruby 4.0 and newer. Ruby 3.3 and 3.4 remain skipped there pending the CRuby backports:

- ruby/ruby#18502 for Ruby 3.3
- ruby/ruby#18503 for Ruby 3.4

## Testing

- `BUNDLE_MIRROR__RUBYGEMS__ORG=https://rubygems.org BUNDLE_GEMFILE=gems.rb bundle exec bake test`
- 541 tests passed with 1,193 assertions on Ruby 4.0.6 using io-event 1.21.0.
- CI passes across CRuby, selector backends, worker-pool coverage, lint, documentation, CodeQL, and external tests.
- The JRuby timeout cascade and TruffleRuby `Fiber.set_scheduler` failure reproduce identically on the previous commit (`57ffab4`) and are unrelated to this update.

## Types of Changes

- Bug fix.

## Contribution

### DIFF
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -426,7 +426,7 @@ module Async
 			while true
 				status = @selector.process_wait(fiber, pid, flags)
 				
-				# `false` indicates the wake-up was spurious, e.g. a stale {unblock}:
+				# `false` indicates the process wait was interrupted before completion:
 				return status unless status == false
 			end
 		end
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -324,9 +324,7 @@ module Async
 				end
 			end
 			
-			# A selector wait may return a falsy result when the fiber is resumed without the requested IO becoming ready.
-			# For example, a deferred unblock from a previous blocking operation may arrive after the fiber has moved on to this wait.
-			# Retry these stale or spurious wake-ups without resetting the original timer.
+			# A selector wait may return a falsy result when the fiber is resumed without the requested IO becoming ready. For example, a deferred unblock from a previous blocking operation may arrive after the fiber has moved on to this wait. Retry these stale or spurious wake-ups without resetting the original timer.
 			until result = @selector.io_wait(fiber, io, events)
 				# If the original timer resumed the fiber, the falsy result represents the timeout rather than a spurious wake-up:
 				return nil if expired
@@ -427,9 +425,7 @@ module Async
 		def process_wait(pid, flags)
 			fiber = Fiber.current
 			
-			# A native process wait may be interrupted before the child exits.
-			# io-event reports this as `false` and leaves the retry policy to the scheduler.
-			# Retry only `false`, since `nil` is a legitimate result for `Process::WNOHANG`.
+			# A native process wait may be interrupted before the child exits. io-event reports this as `false` and leaves the retry policy to the scheduler. Retry only `false`, since `nil` is a legitimate result for `Process::WNOHANG`.
 			while true
 				status = @selector.process_wait(fiber, pid, flags)
 				
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -426,8 +426,12 @@ module Async
 			while true
 				status = @selector.process_wait(fiber, pid, flags)
 				
-				# `false` indicates the wake-up was spurious, e.g. a stale {unblock}
-				return status unless status == false
+				# Native selectors return `false` after a spurious wake-up. Thread-backed waits may return `nil`, but `nil` is also the expected result for a non-blocking wait:
+				if status.nil?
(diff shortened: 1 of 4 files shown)


lib/async/scheduler.rb
```diff
@@ -322,7 +324,13 @@ def io_wait(io, events, timeout = nil)
 				end
 			end
 			
-			return @selector.io_wait(fiber, io, events)
+			loop do
```
ioquatix: It's slightly more efficient to use `while true` for these loops, or better yet, maybe something like `until expired` or something like that.

samuel-williams-shopify:
The final split is now explicit: Async retries only selector-level stale wake-ups (`io_wait` and an explicit `false` from native `process_wait`). It preserves `nil`, including legitimate `Process::WNOHANG` results. The threaded fallback is not worked around in Async or io-event; socketry/io-event#212 adds Ruby 4.0+ regression coverage while ruby/ruby#18502 and ruby/ruby#18503 track the Ruby 3.3/3.4 backports.

ioquatix (approved):

PR #472, opened 2026-08-28
Return false from io_wait() for timeouts

tavianator:
Callers of `rb_io_wait()` expect a `false` return value for timeouts,
not nil.  This could lead to very strange errors from code like

    TCPSocket.new("192.0.2.1", 80, connect_timeout: 1)

which would raise

    TypeError: no implicit conversion from nil to integer

It now raises the correct

    IO::TimeoutError: user specified timeout for 192.0.2.1:80

## Types of Changes

- Bug fix.

## Contribution


samuel-williams-shopify (approved):
Thanks!

PR #473, opened 2026-09-01
Add timeout support to Condition#wait

samuel-williams-shopify:
## Summary

- add a `timeout:` keyword to `Async::Condition#wait`, inherited by `Async::Notification`
- raise `Async::TimeoutError` when the timeout expires
- preserve successful `signal(nil)` by wrapping delivered values internally
- add shared coverage for both condition implementations and an Unreleased note

This allows callers to write `notification.wait(timeout: 1)` directly instead of wrapping the wait with `Task#with_timeout`.

## Validation

- `bundle exec sus` — 546 tests, 1,202 assertions
- RuboCop passes for all changed Ruby files
- `git diff --check`

The repository-wide RuboCop run still reports five pre-existing offenses under `external/`. The Decode documentation coverage task also currently crashes inside Decode while parsing the repository.

### DIFF
--- a/fixtures/async/a_condition.rb
+++ b/fixtures/async/a_condition.rb
@@ -84,6 +84,25 @@ module Async
 			expect(condition.waiting_count).to be == 0
 		end
 		
+		it "can timeout while waiting" do
+			expect do
+				condition.wait(timeout: 0.01)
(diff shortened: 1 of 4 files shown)


ioquatix:
I am not going to do this - better to use `Async::Promise`.

PR #474, opened 2026-09-02
Implement UNIX-style load averages

tavianator:
The existing `Scheduler#load` uses a 1-second time window, which is good
for fast reactivity for things like `Idler`, but less good if you want
it as a metric for health/autoscaling.  Additionally, it performed the
1s time window normalization directly, which has the downside that how
often you called `#load` could change how quickly the load values would
update.

Introduce a 1-minute, 5-minute, and 15-minute time window exponentially-
weighted moving load average, similar to traditional UNIX getloadavg(3).
To avoid expensive math on every fiber switch, we update these averages
at most once a second.  This same 1 second buffer is now used by `#load`
which becomes fully read-only.

## Types of Changes

- New feature.

## Contribution

### DIFF
--- a/lib/async/scheduler.rb
+++ b/lib/async/scheduler.rb
@@ -32,6 +32,13 @@ module Async
 			value == "true" ? true : nil
 		end
 		
+		# The periods over which exponentially weighted load averages are computed, in seconds.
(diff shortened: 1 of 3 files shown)


tavianator:
Going to replace this with something simpler/cheaper

